PR #3, opened 2020-05-15
Sampling is in micro not milli seconds

mrfoto:
1000 milliseconds would be quite a big time chunk 😅


mrfoto:
I signed the CLA but don't know how to re-run the check 🤷‍♂️

PR #4, opened 2020-05-15
Require explicit request authorization in non-development environments

gmcabrita:
This would be a nice to have for production and staging environments.

I took the same approach that rack-mini-profiler does.

An example of authorizing a request for profiling (in Rails):

```ruby
class ApplicationController < ActionController::Base
  before_action do
    if admin_user_signed_in?
      AppProfiler.authorize_request
    end
  end
end
```

### DIFF
--- a/lib/app_profiler.rb
+++ b/lib/app_profiler.rb
@@ -27,6 +27,7 @@ module AppProfiler
   mattr_accessor :logger
   mattr_accessor :root
   mattr_accessor :profile_root
+  mattr_accessor :request_authorization_required
 
   mattr_accessor :speedscope_host, default: "https://speedscope.app"
   mattr_accessor :autoredirect, default: false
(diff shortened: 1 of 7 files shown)


gmcabrita:
The CLA has been signed.

imjching-shopify:
Thanks for contributing! I added `before_profile` and `after_profile` hooks for this use case back then. Would that work for you?

https://github.com/Shopify/app_profiler/blob/80df44c69c8bd7dedb1c250c1ecec422ffa4f838/lib/app_profiler/middleware.rb#L50-L56

To use that, we could create our own middleware with custom logic for those hooks, and specify the custom middleware through `app.config.app_profiler.middleware`. One way to implement this would be to get the controller modify `request.env`, e.g. set `request.env['disallow_profiling']` to `true`, and get `after_profile` to return `false` if `env['disallow_profiling']` is set to `true`.

gmcabrita:
Yeah, that's actually perfect for this use case.

Might be worth documenting this in the README with an example :slightly_smiling_face:

PR #6, opened 2020-05-17
Add access control example to the README

gmcabrita:
Seems like a good idea to document this (following up from #4).

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ You may restrict the storing of profiling results by defining your own Middlewar
 
 For example, the following middleware only stores the profiling results if a `disallow_profiling` key was not added to the `request.env` while processing the request.
 
-```
+```ruby
 class AppProfilerAuthorizedMiddleware < AppProfiler::Middleware
   def after_profile(env, params)
     !env.key?("disallow_profiling")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,6 +71,26 @@ AppProfiler.storage.bucket_name = "new-bucket-name"
 Rails.application.config.app_profiler.storage_bucket_name = "new-bucket-name"
 ```
 
+### Access control
+
+You may restrict the storing of profiling results by defining your own Middleware based on `AppProfiler::Middleware` and changing the `after_profile` hook method to return `false` for such cases.
+
+For example, the following middleware only stores the profiling results if a `disallow_profiling` key was not added to the `request.env` while processing the request.
+
+```
```
imjching, line 80: nit: ` ```ruby `

README.md
```diff
@@ -71,6 +71,26 @@ AppProfiler.storage.bucket_name = "new-bucket-name"
...
+class AppProfilerAuthorizedMiddleware < AppProfiler::Middleware
+  def after_profile(env, params)
+    !env.key?("disallow_profiling")
+  end
+end
+```
+
+The middleware can be enabled by using in the relevant environment configurations:
```
imjching, line 88: I'm not sure if I understand this part properly, but I think this would sound better:
```
The custom middleware can then be configured like the following:
```

PR #11, opened 2020-05-25
Test against Ruby 2.7 and fix warnings

casperisfine:
Fairly straightforward.

I also delete and git ignored the Gemfile.lock. It's better practice for gems as this way you get the newest versions available for the dependencies you declared on each CI run.


.gitignore
```diff
@@ -7,6 +7,8 @@
 /spec/reports/
 /tmp/
 
+/Gemfile.lock
```
Edouard-chin, line 10: Why remove the Gemfile.lock ? It should be under vcs

lib/app_profiler/profiler.rb
```diff
@@ -45,7 +45,7 @@ def start(params = {})
 
         clear
 
-        StackProf.start(DEFAULTS.merge(params))
+        StackProf.start(**DEFAULTS.merge, **params)
```
imjching, line 48: The double-splat operator should be non-destructive and we will always make a copy of the first hash, so I think a merge is unnecessary here. It seems like there was a bug back then that caused the double-splat operator to modify the first hash specified, but it has been patched: https://bugs.ruby-lang.org/issues/9776. With that said, I think this is fine and could help just in case a regression happens (even though very unlikely).

.gitignore
imjching: +1 on this. I think it'll be difficult to track what changed when the lockfile gets removed. There is a possibility in which malicious/bad code (or unintentional breaking change) got introduced in a minor or patch version, and could help as a security precaution.

I see that this was remove to simplify the GitHub tests. Would creating a new step to remove the lockfile before the ruby setup step help solve the initial issue?

casperisfine: As discussed on another gem. Not committing the Gemfile.lock used to be the recommended thing, so when i saw it was extremely outdated I simply 🔥 it.

It's no longer recommended though, so I'll bring it back

casperisfine:
I fixed all comments, please re-review.

lib/app_profiler/profiler.rb
etiennebarrie: The call to merge should be removed here. It fails with ruby 2.5.

casperisfine:
@etiennebarrie the `.merge` shouldn't have been there, it was a total brainfart.

PR #12, opened 2020-06-17
Support a proc to customize the redirect url

DazWorrall:
I'm working to make the configuration of this gem automatic for Shopify Cloud users, but using our custom speedscope installation automatically requires using a non-standard URL. This PR adds a `profile_url_formatter` config option which can be given a proc to allow integrators to specify whatever they want as the redirect location.

I have 🎩 this branch on a real Shopify app and it works as I need.


lib/app_profiler.rb
```diff
@@ -57,5 +58,9 @@ def request_profile_header
     def profile_data_header
       @@profile_data_header ||= profile_header.dup << "-Data" # rubocop:disable Style/ClassVars
     end
+
+    def profile_url_formatter=(prok)
+      @@profile_url_formatter = prok # rubocop:disable Style/ClassVars
```
casperisfine, line 63: Pretty sure class variable is not what you want. Use `@profile_url_formatter` instead.

The above can be cleaned up as well.

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -34,7 +34,11 @@ def append_headers(response, upload:, autoredirect:)
         end
 
         def profile_url(upload)
-          "#{AppProfiler.speedscope_host}#profileURL=#{upload.url}"
+          if AppProfiler.profile_url_formatter.respond_to?(:call)
```
casperisfine, line 37: I'd just do a `nil` check. You expect that property to be a callable, if someone provide something else, it's better if it breaks loudly than to be ignored silently.

test/app_profiler/middleware/upload_action_test.rb
```diff
@@ -68,6 +79,14 @@ def with_autoredirect
       ensure
         AppProfiler.autoredirect = old_autoredirect
       end
+
+      def with_url_formatter(prok)
```
casperisfine, line 83: a more idiomatic name would be `block`

lib/app_profiler.rb
DazWorrall: There's quite a few on this class, should I fix all of them in this PR?

casperisfine: Up to you.

imjching:
I’m thinking about it again, and I don’t think this is the right approach. The profile_url is a concern of the UploadAction, and if one wants to customize that, we should override that, or provide a custom middleware action entirely.

DazWorrall:
>  I’m curious, would overriding the URL methods in the default UploadAction through the app not work for your use case? IIRC, we did this in Core.

This would work for sure, but tbh I did not even look at core - I didn't want to cargo cult it. I'm working backwards from having setup a new app with app_profiler and figuring out the right configuration for Shopify's setup, and now want to make it automatic when used with our platform gems. This kind of integration is the minimum change needed to setup the library using the public API.

> The profile_url is a concern of the UploadAction, and if one wants to customize that, we should override that, or provide a custom middleware action entirely (which we already support).

Overriding works of course, but I wanted to provide a public/supported API.

PR #18, opened 2020-10-07
Generate predictable GCS url

bmansoob:
I often get requests from people (recent one was just now) asking me to "find" their trace in the GCS bucket. If the request times out, if profiler headers are not present in the response, so they cannot find it. It also came up during an incident.

Once this gets shipped, we should update the development guide to let people know how to find their "lost" profile.

### DIFF
--- a/lib/app_profiler/profile.rb
+++ b/lib/app_profiler/profile.rb
@@ -63,14 +63,7 @@ module AppProfiler
     private
 
     def path
-      filename = [
-        Time.zone.now.strftime("%Y%m%d-%H%M%S"),
-        mode,
-        id,
-        Socket.gethostname,
-      ].compact.join("-") << ".json"
-
-      AppProfiler.profile_root.join(filename)
+      AppProfiler.profile_root.join("#{id}.json")
     end
   end
 end


bmansoob:
No clue why lint is failing

```
Error: unrecognized cop Lint/MissingSuper found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml, unrecognized cop Lint/BinaryOperatorWithIdenticalOperands found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
```

imjching:
> No clue why lint is failing
> 
> ```
> Error: unrecognized cop Lint/MissingSuper found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml, unrecognized cop Lint/BinaryOperatorWithIdenticalOperands found in .rubocop-http---shopify-github-io-ruby-style-guide-rubocop-yml
> ```

The `rubocop` gem needs to be bumped: https://github.com/Shopify/ruby-style-guide/commit/d59c1db9808ee9e0b14fdb917259aceb8923283a. It's currently at `0.84.0`.

The root cause is that we are not checking in the rubocop cache here: https://github.com/Shopify/app_profiler/blob/6bd9723c68a2e7929ed93690be403e9644968649/.gitignore#L13-L14

Due to that, it gets pulled all the time whenever the lint operation runs. If the upstream gets updated with backwards incompatible changes, the lint will break. Maybe we should check in the cache into the repository moving forward.

imjching:
I have no strong objections to changing the format, but note that since this is a public gem, we may need to bump its major version since this is not backwards compatible. If I recall correctly, one of the main reasons why the timestamp was included is so that one could easily filter and look for the right file in the GCS bucket through the GCP console.

Making this change also means that we won't be able to tell at a glance whether a particular profile is a `cpu`, `wall`, or `object` profile. I think we might also want to include the `hostname` in the metadata field: https://github.com/Shopify/app_profiler/blob/0e0374100721a89f8f00b02b214069af5a93db0a/lib/app_profiler/request_parameters.rb#L36-L45

If I recall correctly, we might be overriding some of these values in Core: https://github.com/Shopify/app_profiler/blob/5906a30aa2e95633eed5b39d4b03c2ef3dd8124e/lib/app_profiler/middleware/upload_action.rb#L44-L54

Maybe a better approach would be to do this change there instead, i.e. trimming the path part of the URL to only reflect the request ID. Most of the stuff trimmed are already in the `metadata` field as well (There's a custom middleware that does this I think - see `before_profile` or `after_profile` in Core).

bmansoob:
Yup - did not realise this is a public gem 👍 -- will close this and try to fix in Core.

PS: hope you are well!

PR #25, opened 2021-04-21
Assign ownership of app_profiler to sysperf team

dalehamel:
The webscale team no longer exists, and the sysperf team has the most overlap in knowledge
and interests to take on routine maintenance (gem bumps, etc).

cc @honkfestival

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -1,3 +1,10 @@
 classification: library
 slack_channels:
 - webscale
+
+org_line: Production Engineering
+owners:
+  - Shopify/sysperf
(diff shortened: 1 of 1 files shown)


service.yml
```diff
@@ -1,3 +1,10 @@
 classification: library
 slack_channels:
 - webscale
```
honkfestival, line 3: We should update (remove?) this too.

honkfestival (approved):

honkfestival:
:shipit:

PR #30, opened 2021-06-09
Add File Safe Regex to Profile Filename

fresh-eggs:
This PR aims to add a regex which will enforce an allowed list of characters on the profile filenames generated by the App Profiler.

https://github.com/Shopify/app_profiler/blob/4d4ed61eec62e047238c80d0e3e72d256594daa8/lib/app_profiler/profile.rb#L66-L77

The regex `[^0-9A-Za-z.\-\_]`, attempts to match on any character **not** included in alphanumerics, `.`, `-` or `_`.


dalehamel:
thanks @fresh-eggs

PR #33, opened 2021-07-20
Add speedscope remote viewer

gmcgibbon:
Adds remote speedscope viewer for profiling on remote hosts. It uses rack middleware to hijack `/app_profiler` requests to list and view profiles of a booted rack / rails app. Very similar to what rack mini profiler does.

The routes are:

`/app_profiler` to list profiles
`/app_profiler/{id}` to show a specific profile in speedscope
`/app_profiler/viewer/index.html` to mount speedscope via `Rack::File`

### DIFF
--- a/lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
+++ b/lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
@@ -8,8 +8,9 @@ module AppProfiler
       class BaseMiddleware
         class Sanitizer < Rails::Html::SafeListSanitizer
           self.allowed_tags = Set.new(%w(strong em b i p code pre tt samp kbd var sub
-          sup dfn cite big small address hr br div span h1 h2 h3 h4 h5 h6 ul ol li dl dt dd abbr
-          acronym a img blockquote del ins script))
+                                         sup dfn cite big small address hr br div span
+                                         h1 h2 h3 h4 h5 h6 ul ol li dl dt dd abbr
+                                         acronym a img blockquote del ins script))
         end
 
         private_constant(:Sanitizer)
--- a/lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
+++ b/lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
@@ -1,0 +1,98 @@
+# frozen_string_literal: true
+
+require "rails-html-sanitizer"
+
+module AppProfiler
+  module Viewer
+    class SpeedscopeRemoteViewer < BaseViewer
+      class BaseMiddleware
+        class Sanitizer < Rails::Html::SafeListSanitizer
+          self.allowed_tags = Set.new(%w(strong em b i p code pre tt samp kbd var sub
+          sup dfn cite big small address hr br div span h1 h2 h3 h4 h5 h6 ul ol li dl dt dd abbr
+          acronym a img blockquote del ins script))
+        end
+
+        private_constant(:Sanitizer)
+
+        def self.id(file)
+          file.basename.to_s.delete_suffix(".json")
+        end
+
+        def initialize(app)
+          @app = app
+        end
+
+        def call(env)
+          request = Rack::Request.new(env)
+
+          return index(env)                        if request.path_info =~ %r(\A/app_profiler\z)
+          return viewer(env, Regexp.last_match(1)) if request.path_info =~ %r(\A/app_profiler/viewer/(.*)\z)
+          return show(env, Regexp.last_match(1))   if request.path_info =~ %r(\A/app_profiler/(.*)\z)
+
+          @app.call(env)
+        end
+
+        protected
+
+        def id(file)
+          self.class.id(file)
+        end
+
+        def profile_files
+          AppProfiler.profile_root.glob("**/*.json")
+        end
+
+        def render(html)
+          [
+            200,
+            { "Content-Type" => "text/html" },
+            [
+              +<<~HTML,
+                <!doctype html>
+                <html>
+                  <head>
+                    <title>App Profiler</title>
+                  </head>
+                  <body>
+                    #{sanitizer.sanitize(html)}
+                  </body>
+                </html>
+              HTML
+            ],
+          ]
+        end
+
+        def sanitizer
+          @sanitizer ||= Sanitizer.new
+        end
+
+        def viewer(_env, path)
+          raise NotImplementedError
+        end
+
+        def index(_env)
+          render(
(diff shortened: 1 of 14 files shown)


lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
```diff
@@ -0,0 +1,84 @@
...
+            { "Content-Type" => "text/html" },
+            [
+              +<<~HTML,
+                <!doctype html>
+                <html>
+                  <head>
+                    <title>App Profiler</title>
+                  </head>
+                  <body>
+                    #{html}
```
yaworsk, line 47: I believe rendering like this leaves you vulnerable to XSS if `html` ever had anything malicious in it since ActiveSupport provides the default sanitizations

lib/app_profiler/viewer/speedscope_remote_viewer/middleware.rb
```diff
@@ -0,0 +1,58 @@
...
+                var objUrl = encodeURIComponent(URL.createObjectURL(blob));
+                var iframe = document.createElement('iframe');
+
+                document.body.style.margin = '0px';
+                document.body.appendChild(iframe);
+
+                iframe.style.width = '100vw';
+                iframe.style.height = '100vh';
+                iframe.style.border = 'none';
+                iframe.setAttribute('src', '/app_profiler/viewer/index.html#profileURL=' + objUrl + '&title=' + 'Flamegraph for #{name}');
```
yaworsk, line 50: similarly here - not sure if `name` can / will ever be user controlled, but this looks like it would lead to XSS

lib/app_profiler/yarn/command.rb
```diff
@@ -0,0 +1,47 @@
...
+      class YarnError < StandardError; end
+
+      mattr_accessor :yarn_setup, default: false
+
+      def yarn(command)
+        setup_yarn unless yarn_setup
+        exec("yarn #{command}") do
+          raise YarnError, "Failed to run #{command}."
+        end
+      end
```
yaworsk, line 14: This seems very dangerous. Would highly recommend we not just accept any command pass it to yarn. I don't know it well enough but if there's any flag that yarn takes that could result in arbitrary read/write, it would be bad. Would recommend instead we allow-list the commands or actual write the methods with the allowed yarn commands we'd accept

lib/app_profiler/yarn/command.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      def package_json_exists?
+        AppProfiler.root.join("package.json").exist?
+      end
+
+      def exec(command)
+        system(command).tap do |return_code|
+          yield unless return_code
+        end
```
yaworsk, line 43: Similarly, this seems needlessly dangerous. Instead of accepting any command and passing it to `system`, we should allow list what we want to run or just write individual methods with no variability

dalehamel (changes requested):
pending appsec feedback

lib/app_profiler/viewer/speedscope_remote_viewer/base_middleware.rb
gmcgibbon: `sanitize` belongs to Action View which IIRC uses https://github.com/rails/rails-html-sanitizer. I think I can use that here.

lib/app_profiler/viewer/speedscope_remote_viewer/middleware.rb
gmcgibbon: The name would be the match of `\A/app_profiler/(.*)\z` in the path as long as it maps to a file's basename without extension (eg. `tmp/app_profiler/my_profile-4.json` => `my_profile-4`) . Since the filename is sanitized I don't think we need to worry, but I can remove the interpolation and make the title generalized instead if you'd like.

lib/app_profiler/yarn/command.rb
gmcgibbon: This was already in the codebase, I just moved it. Will address in a separate commit.

lib/app_profiler/yarn/command.rb
gmcgibbon: It is a private method and the only consumer is `yarn`. I can inline it if that would make sense to you?

lib/app_profiler/viewer/speedscope_remote_viewer/middleware.rb
NicholasEllul: Where does this sanitization/creation of this filename happen?

NicholasEllul: Thanks!

lib/app_profiler/yarn/command.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def package_json_exists?
+        AppProfiler.root.join("package.json").exist?
+      end
+
+      def exec(*command)
+        system(*command).tap do |return_code|
+          yield unless return_code
+        end
+      end
```
NicholasEllul, line 51: Since we would never want to run `exec` without calling `ensure_command_valid?` can we move the call to that function here? That way down the road if `exec` gets reused somewhere, we can trust that it will always have its safeguards for checking the command bundled into it.

NicholasEllul (commented):
Left a comment to help add more proactive security around the call to `system(...)`. Pete mentioned, the most secure way to use the `system(...)` function is to have an allowlist of things that it can do. You do a great job of this with the `ensure_command_valid?` command, and now the only remaining place of ambiguity is in the `options` parameter.

From what I can see, `yarn init` only gets called with the `--yes` option, and `yarn add` only gets called with the `speedscope` option followed by either the file path, or `--dev --ignore-workspace-root-check`. Narrowing the scope of potential inputs and enforcing devs working on this in the future to be explicit about what they intend to allow will minimize the risk of something slipping between the cracks into this powerful method

lib/app_profiler/yarn/command.rb
gmcgibbon: I changed it to permit all command strings. This should be the most secure option despite the cost of iteration and matching.

NicholasEllul: > hardcode these strings into the methods and forgo allow-listing

Can you expand on what you mean by this? What you've done here is fantastic with the allow list, if we can maintain a similar strict list of commands another way i'm ok with that as well 👍

NicholasEllul (approved):
Left a comment, but i'm happy with the allow list that was added! Re-request a review if theres any changes to that you need me to look over

PR #34, opened 2021-07-22
Clarify storage backend use in development

miharekar:
It took me **waaaaaay** too long to understand why this wasn't working when I was trying to implement an s3 uploader 😅


miharekar:
@gmcabrita thanks! Fixed.

gmcgibbon (approved):

PR #40, opened 2021-12-14
X-Profile-Data value is a string

pedropb:
Some middlewares expect all of the response headers values to be `String` but X-Profile-Data is currently emitting a `Pathname`. This commit change it so that it is explicitly converted to a String, before sending the value downstream.

I ran into an issue specifically with [thin](https://github.com/macournoyer/thin/blob/v1.8.1/lib/thin/response.rb#L63-L77) where it was parsing the response headers and raised the following error:
```
Unexpected error while processing request: undefined method `each' for #<Pathname:0x000000010a196080>
```

Is there a reason to use Pathname over String?

### DIFF
--- a/test/app_profiler/middleware/upload_action_test.rb
+++ b/test/app_profiler/middleware/upload_action_test.rb
@@ -44,6 +44,7 @@ module AppProfiler
 
         assert_predicate(@response[1][AppProfiler.profile_header], :present?)
         assert_predicate(@response[1][AppProfiler.profile_data_header], :present?)
+        assert_equal(@response[1][AppProfiler.profile_data_header].class, String)
         assert_predicate(@response[1]["Location"], :blank?)
         assert_equal(@response[0], 200)
       end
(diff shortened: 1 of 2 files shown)


test/app_profiler/middleware/upload_action_test.rb
```diff
@@ -44,6 +44,7 @@ class UploadActionTest < AppProfiler::TestCase
 
         assert_predicate(@response[1][AppProfiler.profile_header], :present?)
         assert_predicate(@response[1][AppProfiler.profile_data_header], :present?)
+        assert_equal(@response[1][AppProfiler.profile_data_header].class, String)
```
gmcgibbon, line 47:
```suggestion
        assert_kind_of(String, @response[1][AppProfiler.profile_data_header])
```

gmcgibbon (approved):
cc @Shopify/sysperf

pedropb:
FWIW the `Pathname` inside the response header was also breaking [unicorn](https://github.com/defunkt/unicorn/blob/master/lib/unicorn/http_response.rb#L43-L48)

```
ERROR -- : app error: undefined method `=~' for #<Pathname:0x0000041a55ab6fd0> (NoMethodError)
```

PR #44, opened 2022-04-18
Version 0.10.0

swalkinshaw:


swalkinshaw:
🤦  yes, here we go: https://github.com/Shopify/app_profiler/pull/45

PR #47, opened 2022-05-16
log link to speedscope during gcs upload

kim76:
**Why?**

Make it easier to link to the flamegraph in speedscope. When trying to get to the profile I needed to [read through the docs](https://development.shopify.io/engineering/production/performance/profiling) and then eventually disturb @dalehamel to get the right link.

**How?**

- Always output the link to AppProfiler.speedscope_host instead of just the gcs location to make it easier to get to the flame graph and to avoid having to read through the docs
- pulled the link logic that was in the `middleware/upload_action` into the `app_profiler` and accessed the method from `profile`

**Next:** can remove the logging from the [background queue in core](https://github.com/Shopify/shopify/blob/e96c50c572336a434d183a1a383e013f6b709fee/components/platform/app/utils/background_queue/profiler.rb#L25)

**For reviewers:**
- other suggested approaches?


lib/app_profiler/profile.rb
```diff
@@ -41,6 +41,7 @@ def upload
       AppProfiler.storage.upload(self).tap do |upload|
         if upload && defined?(upload.url)
           AppProfiler.logger.info("[Profiler] uploaded profile profile_url=#{upload.url}")
+          AppProfiler.logger.info("[Profiler] Profile available at profile_url=#{AppProfiler.profile_url(upload)}")
```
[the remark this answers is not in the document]
gmcgibbon: I think we're talking about 2 different things in this log message. The profile_url (the location of the profile on the storage service) and the speedscope_url (the address to actually view the profile). I think this would make more sense:

```
AppProfiler.logger.info("[Profiler] uploaded profile profile_url=#{upload.url} speedscope_url=#{AppProfiler.profile_url_formatter.call(upload)}")
```

It would be nice if we could define `#speedscope_url` on the upload object, but this would require a larger refactor of the storage services.

lib/app_profiler.rb
```diff
@@ -72,5 +75,9 @@ def profile_data_header
     def profile_url_formatter=(block)
       @@profile_url_formatter = block # rubocop:disable Style/ClassVars
     end
+
+    def profile_url(upload)
```
gmcgibbon, line 79:
```suggestion
    def profile_viewer_url(upload)
```
If we're going to keep this method, we should name it to distinguish the difference between a profile url and a profile viewer url and use it in `lib/app_profiler/profile.rb` too. `profile_url_formatter` should also be renamed to `profile_viewer_url_formatter`, but we would need to do a deprecation cycle for that (so maybe we could consider doing this in another change).

lib/app_profiler/profile.rb
```diff
@@ -40,7 +40,11 @@ def view
     def upload
       AppProfiler.storage.upload(self).tap do |upload|
         if upload && defined?(upload.url)
-          AppProfiler.logger.info("[Profiler] uploaded profile profile_url=#{upload.url}")
+          AppProfiler.logger.info(
+            "[Profiler] data uploaded: " \
+              "profile_url=#{upload.url} " \
+              "profile_viewer_url=#{AppProfiler.profile_url_formatter.call(upload)}"
+          )
```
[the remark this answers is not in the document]
gmcgibbon:
```suggestion
          AppProfiler.logger.info(
            <<~INFO.squish
              [Profiler] data uploaded:
              profile_url=#{upload.url}
              profile_viewer_url=#{AppProfiler.profile_viewer_url(upload)}
            INFO
          )
```
Nit: We could use a heredoc for this multiline string instead

lib/app_profiler.rb
```diff
@@ -35,7 +35,10 @@ module Viewer
   mattr_accessor :autoredirect, default: false
   mattr_reader   :profile_header, default: "X-Profile"
   mattr_accessor :context, default: nil
-  mattr_reader   :profile_url_formatter, default: nil
+  mattr_reader   :profile_url_formatter,
```
gmcgibbon, line 38:
```suggestion
  mattr_reader   :profile_url_formatter, default: proc do |upload|
    "#{AppProfiler.speedscope_host}#profileURL=#{upload.url}"
  end
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 [Unreleased]: https://github.com/Shopify/app_profiler
+- Log info with link to speedscope during file upload (#47)
```
gmcgibbon, line 10:
```suggestion

- Log info with link to speedscope during file upload (#47)
```

PR #48, opened 2022-05-18
Implement a simple, "HTTP" server and Rack app to initiate profiling remotely

dalehamel:
# TL;DR

This implements a basic, not fully compliant HTTP server, which hosts a minimal rack app that responds to a single endpoint, `/profile`.

This is heavily inspired by [golang's builtin pprof server](https://pkg.go.dev/net/http/pprof), and aims to provide a similar means to passively profiling arbitrary Ruby applications.

# Why

This is implemented to add support for passively collecting profiles from a running application. App profiler already supports profiling individual requests well, but this is intended to profile a whole process, for a specified duration.

In particular, this will be valuable for leveraging stackprof to provide on-CPU and object profiling, which we have no other means to collect easily and passively at the moment.

# Implementation

## Server

A minimal, non-compliant HTTP server is implemented, which does a blocking read on the server socket in a thread, and will process each request in a thread. It is minimal in order to avoid pulling in any additional dependencies, and is built with just the ruby standard library. It is "Rack app compliant-ish". For the intended use here of supporting just a GET request to a single path with some params, this should be sufficient.

The port that it listens on can be specified, but is intended not to be. It is meant to pick a free port, and bind to that so that each process will have a sort of "debug back door" for profiling.

The port will be logged as well as indicated via the name of a tempfile in a specific location, so that it can be easily discovered. This process is documented in the header comment for the file.

This will allow an external process to discover the port to connect to, and then pull ruby profiles of a specified type and duration for the target process by simply making an HTTP request.

Note that for safety / security reasons, it is **must only bind to localhost only**, meaning that it cannot be called be external clients. Otherwise, this could easily be a DOS vector, and information leak. Additionally, in production it is recommended (and defaults) to use a Unix socket, which will **only** work on the localhost.

As it binds on a completely separate socket, this will not be part of any default application router in rails. This also means it will work on any ruby process using app profiler that calls it, including non web server processes like jobs, etc.

The overhead of this server should be essentially nil if it is not called, as it will simply create a socket, and enter a blocking syscall waiting for a request to process. If the application that includes it is a rails app, the railtie will create one instance per process, should the rails app fork, using activesupport's `ForkTracker`.

## Application

The profiling application here is a Rack app. It implements the `/profile` endpoint, and does various error handling to ensure it is only receiving requests it knows how to handle.

It prevents trying to profile while a profile is already running with a semaphore. The semaphore is acquired and checked at the same time, so that other threads using the same class instance cannot initiate profiling if a profile is already running.

The profile mode and duration of the profiling session can both be specified as parameters. If these parameters are invalid, it should return an error to the client.

To profile, it will simply call `AppProfiler.start`, then sleep for the specified duration, then end the profile. This should let it observe what the rest of the VM is doing. The results are then serialized to JSON and returned in the rack response.

### Usage with Speedscope (locally)

CORS is enabled, so you can call the server directly from speedscope provided that you URL encode the address.

For example, with speedscope running locally on :9292, we can have it directly load a profile from the target app using port 57510. The host before URL encoding is: `http://127.0.0.1:57510/profile?duration=1`, and this can be specified to #profileURL so that speedscope will initiate the profiling session by entering `http://127.0.0.1:9292/#profileURL=http%3A%2F%2F127.0.0.1%3A57510%2Fprofile%3Fduration%3D1` in the URL bar.

# Configuration

This can be configured via a railtie. **It defaults to disabled for now**.

Adding the following to the development/production.rb:

```
config.app_profiler.server_enabled = true
```

Results in it being enabled. Readme documents additional config options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,13 @@ Alternatively, the server can be directly started with:
 AppProfiler::Server.start!
 ```
 
-The default duration, for requests without a duration parameter, can also be
+The default duration (in seconds), for requests without a duration parameter, can also be
 set via the railtie config.
 
 ```
-AppProfiler.server.duration
+AppProfiler.server.duration = 30
+# OR
+Rails.application.config.app_profiler.server_duration = 30
 ```
 
 The server supports both TCP and Unix sockets for its transport. It is
@@ -158,13 +160,17 @@ recommended to use TCP for local development, and Unix sockets for production:
 
 ```
 AppProfiler.server.transport = AppProfiler::Server::TRANSPORT_UNIX
+# OR
+Rails.application.config.app_profiler.server_transport = AppProfiler::Server::TRANSPORT_UNIX
 ```
 
 It is possible, but not recommended, to hardcode the listen port to be used in
 TCP server mode with:
 
 ```
-AppProfiler.server.port
+AppProfiler.server.port = 8080
+# OR
+Rails.application.config.app_profiler.server_port = 8080
 ```
 
 If this is done in production and it can cause port conflicts with multiple
@@ -203,14 +209,8 @@ $ echo $SOCK
 
 The API is very simple, and passes supported parameters directly to stackprof.
 
-Supported are:
-
-- mode: the stackprof mode to use [cpu, wall, object]
-- interval: the interval to use
-  - For cpu and wall, this will be the duration in microseconds between samples
-  - For object, this be the modulus of which allocations are counted. Eg, for
-1, every allocation is counted. For 10, only every tenth will be.
-- duration: how long the profiling session should last
+See the [possible keys](#possible-keys) for additional documentation on the
+supported parameters.
 
 For example, to collect a heap profile for 60 seconds, counting every 10th
 allocation:
--- a/README.md
+++ b/README.md
@@ -111,6 +111,129 @@ The custom middleware can then be configured like the following:
 Rails.application.config.app_profiler.middleware = AppProfilerAuthorizedMiddleware
 ```
 
+## Profile Server
+
+This option allows for profiles to be passively collected via an HTTP endpoint,
+inspired by [golang's built-in pprof server](https://pkg.go.dev/net/http/pprof).
+
+A minimal Rack app runs a minimal (non-compliant) HTTP server, which exposes an
+endpoint that allows for profiling. For security purposes, the server is bound
+to localhost only. The HTTP server is built using standard library modules only,
+in order to keep dependencies minimal. Because it is an entirely separate server,
+listening on an entirely separate socket, this should not interfere with any
+existing application routes, and should even be usable in non-web apps.
+
+This allows for two main use cases:
+
+- Passive profile collection in production
+  - Periodically profiling production apps to analyze them responding to real
+workloads
+  - Providing a statistical, long-term view of the "hot paths" of the workload
+- Local development profiling
+  - Can be used to get a profile "on demand" against a development server
+
+### Configuration
+
+If using as a railtie, only a single option needs to be set:
+
+```
+config.app_profiler.server_enabled = true
+```
+
+Alternatively, the server can be directly started with:
+
+```
+AppProfiler::Server.start!
+```
+
+The default duration, for requests without a duration parameter, can also be
+set via the railtie config.
+
+```
+AppProfiler.server.duration
+```
+
+The server supports both TCP and Unix sockets for its transport. It is
+recommended to use TCP for local development, and Unix sockets for production:
+
+```
+AppProfiler.server.transport = AppProfiler::Server::TRANSPORT_UNIX
+```
+
+It is possible, but not recommended, to hardcode the listen port to be used in
+TCP server mode with:
+
+```
+AppProfiler.server.port
+```
+
+If this is done in production and it can cause port conflicts with multiple
+instances of the app, which is another reason why the Unix transport is
+preferred for production.
+
+#### Discovering the port or socket path
+
+In general, the server should be run without setting the port, in which case
+any free TCP port may be used. To determine what the port is, check the
+application logs, or resolve it from the special "Magic file" which contains
+a mapping of pid to port:
+
+```
+$ PID=49825
+$ port_file=$(ls -1 /tmp/app_profiler/profileserver-$PID-port-*)
+$ PORT=$(echo $port_file | sed 's/.*port-\([[:digit:]]*\)-.*/\1/g')
+$ echo $PORT
+60160
+```
+
+This approach is intended to be "machine friendly" so that an external
+profiling agent can easily detect what port to profile on.
+
+For the Unix mode, this is even easier as the file simply includes the PID
+in it, and this will be the file handle to use for the Unix socket:
+
+```
+$ PID=41016
+$ SOCK=$(ls -1d /tmp/app_profiler/* | grep $PID.sock)
+$ echo $SOCK
+/tmp/app_profiler/app-profiler-41016.sock
+```
+
+### Collecting a profile
+
+The API is very simple, and passes supported parameters directly to stackprof.
+
+Supported are:
+
+- mode: the stackprof mode to use [cpu, wall, object]
+- interval: the interval to use
+  - For cpu and wall, this will be the duration in microseconds between samples
+  - For object, this be the modulus of which allocations are counted. Eg, for
+1, every allocation is counted. For 10, only every tenth will be.
+- duration: how long the profiling session should last
+
+For example, to collect a heap profile for 60 seconds, counting every 10th
+allocation:
+
+```
+curl "http://127.0.0.1:$PORT/profile?duration=60&mode=object&interval=10"
+```
+
+#### Usage with speedscope directly
+
+By default the server will allow CORS. This can be disabled if it presents a
+problem, but it should be generally safe given that the server listens for
+requests on localhost only, which is already a private network address.
+
+This can be used with a local instance of speedscope to directly initiate
+profiling from the browser. Assuming speedscope is running locally on port
+`9292`, and the profile server is running on port `57510`, the server address
+can  be URL encoded, and passed to speedscope via `#profileURL`:
+
+```
+http://127.0.0.1:9292/#profileURL=http%3A%2F%2F127.0.0.1%3A57510%2Fprofile%3Fduration%3D1
+```
+
 ## Profiling manually
 
 `AppProfiler` can be used more simply to profile blocks of code. Here's how:
--- a/lib/app_profiler/railtie.rb
+++ b/lib/app_profiler/railtie.rb
@@ -45,9 +45,9 @@ module AppProfiler
 
     initializer "app_profiler.enable_server" do
       if AppProfiler.server.enabled
-        AppProfiler::Server.start!
+        AppProfiler::Server.start
         ActiveSupport::ForkTracker.after_fork do
-          AppProfiler::Server.start!
+          AppProfiler::Server.start
         end
       end
     end
--- a/lib/app_profiler/railtie.rb
+++ b/lib/app_profiler/railtie.rb
@@ -17,6 +17,7 @@ module AppProfiler
       AppProfiler.middleware.action = app.config.app_profiler.middleware_action || default_middleware_action
       AppProfiler.middleware.disabled = app.config.app_profiler.middleware_disabled || false
       AppProfiler.server.enabled = app.config.app_profiler.server_enabled || false
+      AppProfiler.server.transport = app.config.app_profiler.server_transport || default_appprofiler_transport
       AppProfiler.server.port = app.config.app_profiler.server_port || 0
       AppProfiler.server.duration = app.config.app_profiler.server_duration || 30
       AppProfiler.server.cors = app.config.app_profiler.server_cors || true
@@ -53,6 +54,16 @@ module AppProfiler
         Middleware::ViewAction
       else
         Middleware::UploadAction
+      end
+    end
+
+    def default_appprofiler_transport
+      if Rails.env.development?
+        # default to TCP server in development so that if wanted users are able to target
+        # the server with speedscope
+        AppProfiler::Server::TRANSPORT_TCP
+      else
+        AppProfiler::Server::TRANSPORT_UNIX
       end
     end
   end
--- a/lib/app_profiler/railtie.rb
+++ b/lib/app_profiler/railtie.rb
@@ -44,7 +44,12 @@ module AppProfiler
     end
 
     initializer "app_profiler.enable_server" do
-      AppProfiler::Server.start! if AppProfiler.server.enabled
+      if AppProfiler.server.enabled
+        AppProfiler::Server.start!
+        ActiveSupport::ForkTracker.after_fork do
+          AppProfiler::Server.start!
+        end
+      end
     end
 
     private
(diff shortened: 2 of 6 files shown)


lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+    # It will create an extremely minimal rack environment hash and hand it off
+    # to our application to process
+    class ProfileServer
+      PROFILER_TEMPFILE_PATH = "/tmp/app_profiler" # for tempfile that indicates port in filename or unix sockets
+
+      class ProfileServerTransport
+        attr_reader :socket
+
+        def client
+          raise("'client' method not implemented for transport")
```
gmcgibbon, line 135:
```suggestion
          raise(NotImplementedError)
```

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+        end
+
+        def stop
+          raise("'stop' method not implemented for transport")
```
gmcgibbon, line 139:
```suggestion
          raise(NotImplementedError)
```
You can add the text, but I think it is straightforward enough without it.

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+      end
+    end
+
+    # This is a minimal, non-compliant "HTTP" server.
+    # It will create an extremely minimal rack environment hash and hand it off
+    # to our application to process
+    class ProfileServer
+      PROFILER_TEMPFILE_PATH = "/tmp/app_profiler" # for tempfile that indicates port in filename or unix sockets
+
+      class ProfileServerTransport
```
gmcgibbon, line 131:
```suggestion
      class Transport
```
This is already within the `ProfileServer` module, so `ProfileServer::Transport` makes more sense to me. It also allows for more readable subclasses: `class UNIX < Transport`.

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+
+      def stop
+        @listen_thread.kill
+        @transport.stop
+      end
+    end
+
+    class << self
+      def reset!
+        @profile_server = {}
```
gmcgibbon, line 258:
```suggestion
        profile_server.clear
```
might be more efficient.

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+      def stop!
+        return unless profile_server.key?(Process.pid)
+
+        profile_server[Process.pid].stop
+        profile_server.delete(Process.pid)
+      end
+
+      private
+
+      def profile_server
```
gmcgibbon, line 287: Nit: Generally I would expect `profile_servers` to equal `{}` and `profile_server` to be `profile_servers[Process.pid]`. Would the hash here actually be shared among processes though? What is the scenario where we expect there to be more than one server?

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+
+    class << self
+      def reset!
+        @profile_server = {}
+        DEFAULTS.each do |config, value|
+          class_variable_set(:"@@#{config}", value) # rubocop:disable Style/ClassVars
+        end
+      end
+
+      def start!
```
gmcgibbon, line 264: Personally, I don't think `stop!` and `reset!` need to have exclamation marks. Usually that is used to denote a dangerous alternative to a method and not that the method mutates itself.

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+      end
+
+      def serve
+        return unless @listen_thread.nil?
+
+        app = ProfileApplication.new
+
+        @listen_thread = Thread.new do
+          loop do
+            Thread.new(@transport.socket.accept) do |session|
```
gmcgibbon, line 212: This will spin up a bunch of threads in the loop. Is this intended, or should the loop be inside this block?

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+        if request.request_method != "GET"
+          response.status = HTTP_NOT_ALLOWED
+          response.write("Only GET requests are supported")
+          return response
+        end
+        case request.path
+        when "/profile"
+          begin
+            stackprof_args, duration = validate_profile_params(request.params)
+          rescue => e
```
gmcgibbon, line 62: This is a catch-all which could result in unexpected return messages. Maybe we should be specific about what error we expect to catch here?

lib/app_profiler/server.rb
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    mattr_accessor :enabled, default: DEFAULTS[:enabled]
+    mattr_accessor :transport, default: DEFAULTS[:transport]
+    mattr_accessor :cors, default: DEFAULTS[:cors]
+    mattr_accessor :cors_host, default: DEFAULTS[:cors_host]
+    mattr_accessor :port, default: DEFAULTS[:port]
+    mattr_accessor :duration, default: DEFAULTS[:duration]
+
+    class ProfileApplication
```
gmcgibbon, line 37: Do we want children of the server module to be public or private API? Perhaps we should use `private_constant` to make some of this private.

lib/app_profiler/railtie.rb
```diff
@@ -37,6 +43,15 @@ class Railtie < Rails::Railtie
       end
     end
 
+    initializer "app_profiler.enable_server" do
+      if AppProfiler.server.enabled
+        AppProfiler::Server.start!
+        ActiveSupport::ForkTracker.after_fork do
```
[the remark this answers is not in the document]
gmcgibbon: Makes sense, but do we need to start the server on the main thread as well as forks?

Also we may need a require for the fork tracker if it isn't always loaded by default.

README.md
```diff
@@ -111,6 +111,129 @@ The custom middleware can then be configured like the following:
...
+$ SOCK=$(ls -1d /tmp/app_profiler/* | grep $PID.sock)
+$ echo $SOCK
+/tmp/app_profiler/app-profiler-41016.sock
+```
+
+### Collecting a profile
+
+The API is very simple, and passes supported parameters directly to stackprof.
+
+Supported are:
```
gmcgibbon, line 206: Can we just link to the section above this that explains the possible options instead of reiterating them?

README.md
```diff
@@ -111,6 +111,129 @@ The custom middleware can then be configured like the following:
...
+
+This option allows for profiles to be passively collected via an HTTP endpoint,
+inspired by [golang's built-in pprof server](https://pkg.go.dev/net/http/pprof).
+
+A minimal Rack app runs a minimal (non-compliant) HTTP server, which exposes an
+endpoint that allows for profiling. For security purposes, the server is bound
+to localhost only. The HTTP server is built using standard library modules only,
+in order to keep dependencies minimal. Because it is an entirely separate server,
+listening on an entirely separate socket, this should not interfere with any
+existing application routes, and should even be usable in non-web apps.
```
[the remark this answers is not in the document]
dalehamel: I have considered this and flip-flopped on it a fair bit, but at least for now I think this gem is the best spot for it.

This functionality provides a new way to tie into the app profiler and ultimately profile apps, it just doesn't focus on individual requests, and is concerned with profiling at the process level. This can be a good starting point for profiling an app, which would lead to profiling specific requests, so I think the functionality is still pretty complementary.

Since the functionality is mostly in one new file, I don't think it would be too hard to separate in the future, but for now I think a separate gem would be overkill for this one feature, and worry about the overhead it would add it managing dependencies and rolling this out. So for now, I think just putting it behind a configuration flag in the existing gem makes sense.

Any thoughts on what a "tripwire" for deciding to move to a separate gem would be?

README.md
```diff
@@ -111,6 +111,129 @@ The custom middleware can then be configured like the following:
...
+
+This allows for two main use cases:
+
+- Passive profile collection in production
+  - Periodically profiling production apps to analyze them responding to real
+workloads
+  - Providing a statistical, long-term view of the "hot paths" of the workload
```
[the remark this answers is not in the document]
dalehamel: "Real workloads" refers to that this profiles the actual requests an application is receiving organically, ie, not a specific endpoint that was asked to be profiled, which is somewhat a "synthetic" request.

"Hot paths" indicates what parts of the code are showing up in the most stack traces, ie, regardless of what request is being performed. This might indicate a specific request is slow, or that a shared path used by different related requests is slow, etc.

Does that help to clarify?

README.md
dalehamel: Good idea 👍

lib/app_profiler/railtie.rb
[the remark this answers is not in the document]
dalehamel: > do we need to start the server on the main thread as well as forks

In many cases we won't need to, but this is to cover the case where the server / target process does **not** fork. There is no harm in running it in the case of a puma / unicorn master process anyways, so the approach is a bit of a "catch-all".

> Also we may need a require for the fork tracker if it isn't always loaded by default.

Good call, I'll add that, thanks!

lib/app_profiler/server.rb
dalehamel: thanks, i'll amend this!

lib/app_profiler/server.rb
dalehamel: good point, i'll update this 👍

dalehamel: Updated and added a comment to clarify the reason for the hash

lib/app_profiler/railtie.rb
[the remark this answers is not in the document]
gmcgibbon: Sorry, if it is autoloaded we don't need a require. We can assume `require "active_support"`

README.md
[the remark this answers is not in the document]
gmcgibbon: Yes, but I think it would help me to see what kind of profile data this produces in comparison to the traditional approach of profiling a request of a Rails app. Since the thread just sleeps, I assumed it would just be a blank profile.

test/app_profiler/profile_server_test.rb
```diff
@@ -0,0 +1,332 @@
...
+        Server.reset
+        refute_equal(TRANSPORT_TCP, Server.transport)
+        assert_equal(DEFAULTS[:transport], Server.transport)
+      end
+    end
+
+    class TCPServerTest < TestCase
+      def setup
+        Server.reset
+        Server.transport = TRANSPORT_TCP
```
gmcgibbon, line 28: Do we need to revert this assignment in this test case or the one below in a teardown? I don't think test cases should finish and leave global state changed. You can do something like

```
def with_transport(value)
  old_transport = Server.transport
  Server.transport = value
  yield
ensure
  Server.transport = old_transport
end

def run(*)
  with_transport(SOME_CONSTANT) do
    super
  end
end
```
or just use a teardown block.

lib/app_profiler/railtie.rb
```diff
@@ -37,6 +43,16 @@ class Railtie < Rails::Railtie
       end
     end
 
+    initializer "app_profiler.enable_server" do
+      if AppProfiler.server.enabled
+        require "active_support/fork_tracker"
```
gmcgibbon, line 48: We can remove this. Sorry, my mistake.

README.md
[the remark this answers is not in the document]
dalehamel: > what kind of profile data this produces in comparison to the traditional approach of profiling a request of a Rails app

TL;DR, it will provide for aggregated / averaged profiles (see the examples at the end), displayed in a format similar to the "left heavy" view in speedscope. It will show an aggregate view of **everything** the app is doing during these profiling sessions.

>  Since the thread just sleeps, I assumed it would just be a blank profile.

The sleeping thread shouldn't show up much, it should predominantly show threads that are actually doing work, ie, 
 this will profile the **other threads** in the application, though more generally, it will get the stack trace for "whatever is running at the time". My understanding of stackprof is that:

- For CPU and Wall modes, an `itimer` is set (based on the configured "interval") then it will get whatever stack is the "active" one in the ruby VM at the time (stackprof itself works via interrupts / tracepoints, so doesn't have its own "thread", which is why it itself never shows up)
  - For Wall mode, the timer is based on real time elapsing
  - For CPU mode, the timer should only be "charged" when a thread is actually running
- For Object mode, it will be based on a certain number of allocations (identified by the sample rate) Eg, "1" would be for every object allocation, for "1000", it would be every thousandth allocation, etc.

Note that it is still possible to see "sleep" in the CPU profile, because a sleeping thread in ruby will [periodically awaken, then return to a native sleep](https://github.com/ruby/ruby/blob/master/thread.c#L1264), and this might coincide with the timer expiring given enough samples.

For Object mode, it is based on tracepoints and not a timer, so we will only see the stacks for other threads which are actually allocating objects (we won't see the profile server, since it is sleeping and therefore not allocating anything).

I think the most interesting data here is the CPU and object allocation, which should tell you:

- CPU: Where is your application actually spending its CPU time when it gets a chance to run?
- Object: What code paths in your application are responsible for allocating objects (how many?)

Here are some samples of real data collected (albeit, against a relatively idle test app) with this, displayed in cloud profiler. The samples were collected approximately once a minute, for approximately 10 seconds. These should be the equivalent of the "left heavy" view in speedscope.

CPU graph (basically showing what an idle server uses its CPU time for, not terribly exciting, would be more interesting if we had some requests happening, but this is a good "baseline"):

Screen Shot 2022-06-14 at 2 00 15 PM


Object profile graph:

Screen Shot 2022-06-14 at 2 01 45 PM

The object profile graph shows the count of objects allocated on average by each stack frame, eg, the selected frame in the UI here had ~ 4 objects on average.

Note that both of these graphs are the aggregate of many profiling sessions, each representing the average of ~ 250 sessions. This thus gives us a pretty good average view of what threads are actually doing work in the application, and where allocations are happening. The larger the number of samples, the better the view.

Again, of course, the data would be more interesting for an active app, as we are seeing the CPU graph dominated by threads periodically waking up from their sleep (including our own profile server), and a relatively small number of object allocations.

Does this help to clarify a bit? This is a relatively novel use case for ruby profiling both at Shopify and I think in general, but exists for other languages (eg, golang as I have mentioned), so I believe it will be useful for developers once we expose it, as this sort of data has been used for Go apps for a few years now.

test/app_profiler/profile_server_test.rb
dalehamel: good point, i have been bit by this more than once already, I think doing this in a teardown block would be ideal (along with your superclass suggestion above too, I think).

dalehamel:
@gmcgibbon I've taken your suggestions to clean up the test suite, thanks for those. I believe everything should be addressed now.

I've also updated the changelog, as I'd like to follow this up with a new release once this merges.

If you'd like to give it one more look before I merge let me know, otherwise let me know if you think this is good to ship (I see you already approved, thanks), and i'll merge and start the release process.

lib/app_profiler/railtie.rb
```diff
@@ -37,6 +43,15 @@ class Railtie < Rails::Railtie
       end
     end
 
+    initializer "app_profiler.enable_server" do
+      if AppProfiler.server.enabled
```
rafaelfranca, line 47: Should this server starts when running tests, running migrations or starting the console? Or only when the Rails servers starts? Right now it starts every time.

dalehamel: It should be enabled for any long-running process, so rails server or jobs.

What is the most idiomatic way to achieve this?

PR #61, opened 2022-08-16
[Profile Server] only require neccesary files

bmansoob:
Forward fix for the comment here: https://github.com/Shopify/app_profiler/pull/59#discussion_r947059314. Thanks, Rafael.


bmansoob:
> FYI when i did a minimal test it looks like we also need to explicitly require `json` @bmansoob

thanks -- added.

PR #62, opened 2022-08-16
Require only needed core extensions and 'json' for server

dalehamel:
Per your comment https://github.com/Shopify/app_profiler/pull/59/files#r947059314 @rafaelfranca 

I figured I would fix forward rather than revert if that's ok.

I tested with a minimal ruby script and these seem to be the only required extensions. We had also missed requiring `json`, so I added that explicitly as well.


dalehamel:
Duplicate of #61

PR #63, opened 2022-09-19
Railtie is optional dependency

kirs:
In a PR like https://github.com/Shopify/storefront-renderer/pull/15043 when you make `app_profiler` a dependency of a non-Rails app, it would still bring `railties` as a dependency. But that's not necessary and we could make Railtie an optional dependency just for Rails apps.

### DIFF
--- a/app_profiler.gemspec
+++ b/app_profiler.gemspec
@@ -28,7 +28,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency("activesupport", ">= 5.2")
   spec.add_dependency("rack")
-  spec.add_dependency("railties", ">= 5.2")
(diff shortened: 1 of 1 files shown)


dalehamel:
I'm guessing this shouldn't cause problems for the majority of Ruby apps, as they are already rails apps and will already have railties as a dependency? And if they aren't a rails app, it should likewise be fine because the code simply won't be loaded?

alexcoco:
> And if they aren't a rails app, it should likewise be fine because the code simply won't be loaded?

Having these dependencies creates extra load on the team to bump dependencies and understand how they are used within the app. If the dependencies are not needed I would prefer they were not dependencies at all.

alexcoco (approved):

PR #64, opened 2022-09-21
Take logger as an arg in server start method

bmansoob:
Currently `AppProfiler::Server` is using `AppProfiler.logger` which refers `Rails`. This runs into Sorbet issues for apps that are not Rails based (internally, SFR).

After discussing it with @dalehamel and @kirs -- we are taking logger as an arg for the `AppProfiler::Server#start` method. 

Since this is a breaking change, we have added `null` logger as the default, so at least it does not break existing clients. For apps that are using Rails, we are setting up this correctly in `Railtie`. For a small amount of users we will lose logs, if they do not pass it in. I think this is a reasonable compromise.

CC: @ahayworth

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,6 @@ PATH
     app_profiler (0.1.2)
       activesupport (>= 5.2)
       rack
-      railties (>= 5.2)
       stackprof (~> 0.2)
 
 GEM
--- a/test/app_profiler/profile_server_test.rb
+++ b/test/app_profiler/profile_server_test.rb
@@ -8,7 +8,10 @@ module AppProfiler
     TEST_PORT = 11337
 
     class ProfileServerTestCase < TestCase
-      setup { Server.reset }
+      setup do
+        @logger = Logger.new(IO::NULL)
+        Server.reset
+      end
       teardown { Server.reset }
     end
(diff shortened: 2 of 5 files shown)


Gemfile.lock
```diff
@@ -4,7 +4,6 @@ PATH
     app_profiler (0.1.2)
       activesupport (>= 5.2)
       rack
-      railties (>= 5.2)
```
bmansoob, line 7: This is probably due to https://github.com/Shopify/app_profiler/pull/63 and not the changes here.

test/app_profiler/profile_server_test.rb
```diff
@@ -8,7 +8,10 @@ module Server
     TEST_PORT = 11337
 
     class ProfileServerTestCase < TestCase
-      setup { Server.reset }
+      setup do
+        @logger = Logger.new("/dev/null")
```
bmansoob, line 12: probably should use `IO::NULL` abstraction, will change.

dalehamel (approved):
Looks good, thanks for the great PR description.

kirs:
I can confirm this fixes the problem we've had in storefront-renderer 👌

kirs (approved):

PR #80, opened 2023-06-07
Allow out-of-band, async upload functionality

bmansoob:
TL;DR; Adds a new query string param `async`. When specified, profile uploads happen in a background thread. 

One use case for this is when we want to profile a certain % of traffic without end users explicitly requesting it. In such situations uploading profiles inline does not scale.

From the `README`:

>If the `async` key in the query string is set to `1` or `true`, then profile will be uploaded later, in an async manner. One use case for this is when we want to profile a certain % of traffic without incurring costs of inline profile uploads.

If the `async` parameter is specified we add to a queue/buffer -- which we process every `upload_queue_interval_secs`.  The queue is [SizedQueue](https://ruby-doc.org/core-2.5.0/SizedQueue.html). 

CC @jasonhl

### DIFF
--- a/lib/app_profiler/middleware/upload_action.rb
+++ b/lib/app_profiler/middleware/upload_action.rb
@@ -4,16 +4,19 @@ module AppProfiler
   class Middleware
     class UploadAction < BaseAction
       class << self
-        def call(profile, response: nil, autoredirect: nil)
-          profile_upload = profile.upload
+        def call(profile, response: nil, autoredirect: nil, async: false)
+          if async
+            profile.enqueue_upload
+            response[1][AppProfiler.profile_async_header] = true
+          else
+            profile_upload = profile.upload
 
-          return unless response
-
-          append_headers(
-            response,
-            upload: profile_upload,
-            autoredirect: autoredirect.nil? ? AppProfiler.autoredirect : autoredirect
-          )
+            append_headers(
+              response,
+              upload: profile_upload,
+              autoredirect: autoredirect.nil? ? AppProfiler.autoredirect : autoredirect
+            ) if response
+          end
         end
 
         private
--- a/lib/app_profiler/railtie.rb
+++ b/lib/app_profiler/railtie.rb
@@ -28,11 +28,14 @@ module AppProfiler
         "APP_PROFILER_SPEEDSCOPE_URL", "https://speedscope.app"
       )
       AppProfiler.profile_header = app.config.app_profiler.profile_header || "X-Profile"
+      AppProfiler.profile_async_header = app.config.app_profiler.profile_async_header || "X-Profile-Async"
       AppProfiler.profile_root = app.config.app_profiler.profile_root || Rails.root.join(
         "tmp", "app_profiler"
       )
       AppProfiler.context = app.config.app_profiler.context || Rails.env
       AppProfiler.profile_url_formatter = app.config.app_profiler.profile_url_formatter
+      AppProfiler.upload_queue_max_length = app.config.app_profiler.upload_queue_max_length || 10
+      AppProfiler.upload_queue_interval_secs = app.config.app_profiler.upload_queue_interval_secs || 5
     end
 
     initializer "app_profiler.add_middleware" do |app|
--- a/lib/app_profiler/request_parameters.rb
+++ b/lib/app_profiler/request_parameters.rb
@@ -12,6 +12,10 @@ module AppProfiler
       query_param("autoredirect") || profile_header_param("autoredirect")
     end
 
+    def async
+      query_param("async")
+    end
+
     def valid?
       if mode.blank?
         return false
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ Profiling can be triggered in one of two ways:
    - `X-Profile` header format: `[<key>=<value>];...`
 
 
+If `async` query string is provided, then the profile will be uploaded later, in an async manner. One use case for this is when we want to profile a certain % of traffic without incurring costs of inline profile uploads.
+
 You can configure the profile header using:
 
 ```ruby
@@ -49,6 +51,7 @@ Rails.application.config.app_profiler.profile_header = "X-Profile"
 | Key | Value | Notes |
 | --- | ----- | ----- |
 | profile/mode | Supported profiling modes: `cpu`, `wall`, `object`. | Use `profile` in (1), and `mode` in (2). |
+| async | Upload profile in a background thread. When this is set, profile redirect headers are not present in the response.
 | interval | Sampling interval in microseconds. | |
 | ignore_gc | Ignore garbage collection frames | |
 | autoredirect | Redirect request automatically to Speedscope's page after profiling. | |
--- a/lib/app_profiler.rb
+++ b/lib/app_profiler.rb
@@ -39,6 +39,7 @@ module AppProfiler
   mattr_accessor :speedscope_host, default: "https://speedscope.app"
   mattr_accessor :autoredirect, default: false
   mattr_reader   :profile_header, default: "X-Profile"
+  mattr_accessor :profile_async_header, default: "X-Profile-Async"
   mattr_accessor :context, default: nil
   mattr_reader   :profile_url_formatter,
     default: DefaultProfileFormatter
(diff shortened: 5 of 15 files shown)


lib/app_profiler/middleware/upload_action.rb
```diff
@@ -4,9 +4,14 @@ module AppProfiler
   class Middleware
     class UploadAction < BaseAction
       class << self
-        def call(profile, response: nil, autoredirect: nil)
-          profile_upload = profile.upload
+        def call(profile, response: nil, autoredirect: nil, async: false)
+          if async
+            enqueue_upload(profile)
+            response[1][AppProfiler.profile_async_header] = true
```
dalehamel, line 10: response[1] is a rack convention right? We don't need to index bound check it? What about nil check?

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -16,8 +21,43 @@ def call(profile, response: nil, autoredirect: nil)
...
+              sleep(AppProfiler.upload_queue_interval_secs)
+            end
+          end
+          @process_queue_thread.priority = -1 # low priority
+        end
+
         private
 
+        def init_queue
+          @queue = SizedQueue.new(AppProfiler.max_upload_queue_length)
```
dalehamel, line 46: In process_queue we call `init_queue` unconditionally, so could we possibly race here? Ie, could something add something to the queue, then we call init queue and clobber that object? Perhaps this should be an `||=`

dalehamel (commented):
I few questions and a couple nitpicks. Looking great overall. I'll try and think of other tests we might want.

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -16,8 +21,43 @@ def call(profile, response: nil, autoredirect: nil)
...
+        end
+
+        def process_queue
+          return 0 if @queue.nil? || @queue.empty?
+
+          queue = @queue
```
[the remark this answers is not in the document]
bmansoob: discussed this offline - but TL;DR; the need for a local var is to make the "critical section" as small as we can. It was missing a `mutex` which I have added now.

lib/app_profiler/middleware/upload_action.rb
bmansoob: related to https://github.com/Shopify/app_profiler/pull/80#discussion_r1223435346 -- we want to reset it every time we process the queue.

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -16,8 +21,54 @@ def call(profile, response: nil, autoredirect: nil)
...
+        end
+
+        def start_process_queue_thread
+          @process_queue_thread ||= Thread.new do
+            loop do
+              process_queue
+              sleep(AppProfiler.upload_queue_interval_secs)
+            end
+          end
+          @process_queue_thread.priority = -1 # low priority
```
bmansoob, line 42: Not exactly sure if this is needed -- but the idea was to give a low priority from the main thread. https://ruby-doc.org/core-2.5.0/Thread.html#method-i-priority

lib/app_profiler/railtie.rb
```diff
@@ -41,6 +44,9 @@ class Railtie < Rails::Railtie
           app.middleware.insert_before(0, Viewer::SpeedscopeRemoteViewer::Middleware)
         end
         app.middleware.insert_before(0, AppProfiler.middleware)
+        ActiveSupport::ForkTracker.after_fork do
+          AppProfiler::Middleware::UploadAction.start_process_queue_thread
+        end
```
gmcgibbon, line 49: This starts the thread for all usage contexts of the profiler, which I think might be incorrect. If the developer never wants to upload asynchronously, this thread has no purpose. Is it possible to do this lazily?

lib/app_profiler/request_parameters.rb
```diff
@@ -16,6 +16,11 @@ def autoredirect
       query_param("autoredirect") || profile_header_param("autoredirect")
     end
 
+    def async
+      val = query_param("async")
+      val == "true" || val == "1"
```
gmcgibbon, line 21: We typically don't do strict value checking like this. Is there a reason we want only 1 or true and not _any_ truthy value?

robertlaurin:
Curious if you had considered the [Rack Events API](https://www.rubydoc.info/gems/rack/Rack/Events)to defer the upload work to after the request was complete?

There's some folks at Github exploring this approach as an alternative/successor to the OpenTelemetry rack middleware https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/342 so that we can defer more work to outside of the request context.

yimmee123 (approved):

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -16,8 +21,54 @@ def call(profile, response: nil, autoredirect: nil)
...
+            @queue ||= init_queue
+            begin
+              @queue.push(profile, true) # non-blocking push, raises ThreadError if queue is full
+            rescue ThreadError
+              AppProfiler.logger.info("[AppProfiler] upload queue is full, profile discarded")
+            end
+          end
+        end
+
+        def start_process_queue_thread
```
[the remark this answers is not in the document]
bmansoob: I had originally wrote something like [this]

(https://github.com/Shopify/app_profiler/pull/80/commits/fc46ca67713793da5e2f8c5f2ad9260ea8ebb411) but thought it is getting complex so changed it. 

But TL;DR; the queue stuff moved to GCS storage class, new `enqueue_upload` method added to `BaseStorage`. IMO we probably do not need to introduce another abstraction for queue management but let me know if you think differently. 

I think I addressed your other comments too in this commit too.

lib/app_profiler/railtie.rb
bmansoob: This was a good call out, thanks 👍

lib/app_profiler/middleware/upload_action.rb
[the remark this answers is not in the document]
gmcgibbon: Make sense to me 👍

lib/app_profiler/middleware/upload_action.rb
```diff
@@ -4,9 +4,14 @@ module AppProfiler
   class Middleware
     class UploadAction < BaseAction
       class << self
-        def call(profile, response: nil, autoredirect: nil)
-          profile_upload = profile.upload
+        def call(profile, response: nil, autoredirect: nil, async: false)
+          if async
+            profile.enqueue_upload
+            response[1][AppProfiler.profile_async_header] = true
+            return
```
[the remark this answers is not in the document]
bmansoob: done

bmansoob:
Rebased and squashed the commit. I will 🎩 this against core and make sure nothing breaks.

lib/app_profiler/railtie.rb
```diff
@@ -28,11 +28,14 @@ class Railtie < Rails::Railtie
         "APP_PROFILER_SPEEDSCOPE_URL", "https://speedscope.app"
       )
       AppProfiler.profile_header = app.config.app_profiler.profile_header || "X-Profile"
+      AppProfiler.profile_async_header = app.config.app_profiler.profile_async_header || "X-Profile-Async"
       AppProfiler.profile_root = app.config.app_profiler.profile_root || Rails.root.join(
         "tmp", "app_profiler"
       )
       AppProfiler.context = app.config.app_profiler.context || Rails.env
       AppProfiler.profile_url_formatter = app.config.app_profiler.profile_url_formatter
+      AppProfiler.upload_queue_max_length = app.config.upload_queue_max_length || 10
```
bmansoob, line 37: this and the next one are wrong, it should be `app.config.app_profiler....` missed `app_profiler`

bmansoob:
🎩 'ed on an app

Header:

Screenshot 2023-07-27 at 1 47 49 PM

Logs for queue length:

```
cat logs | grep Queue                                                                                                               
Queue size is: 0
Queue size is: 0
Queue size is: 1
Queue size is: 0
Queue size is: 0
Queue size is: 1
Queue size is: 0
```

dalehamel (approved):
Thanks for this 🙏

bmansoob:
Rebased and resolved the merge conflicts.

@casperisfine -- Can you please give this a read, it is somewhat relevant to the work you have done in https://github.com/Shopify/app_profiler/pull/87.

casperisfine (approved):
Implementation looks fine to me. I particularly appreciate the use of a SizedQueue to avoid piling work faster it's dequeued. 

I'm not 100% sure I understand how it's meant to be used though. I see how it could be useful to reduce the impact of profiles triggered programaticaly, but I don't get the need for the `X-Profile-Async` header.

PR #93, opened 2023-09-25
Properly close inherited servers on fork

casperisfine:
I noticed some UNIXServer and AppServer accumulating with reforking enabled.

The reason is that the `at_exit { stop }` make `Server` instances immortal, so they are never GC, and their open socket accumulate and are never closed.

That callback is useless anyway, as on exit the finalizer of these objects will be invoked, and they'll cleanup after themselves.


casperisfine:
> That callback is useless anyway, as on exit the finalizer of these objects will be invoked, and they'll cleanup after themselves.

My bad on that one. `UNIXServer` doesn't delete the file. So I replaced the `at_exit` by an object finalizer.

casperisfine:
> sorry for this oversight.

No worries, it's a common footgun.

PR #94, opened 2023-10-16
Add support custom file prefix block

manuelfelipe:
Move default file prefix behaviour out into a default block and allow for gem consumers to provide a custom block via `AppProfiler.profile_file_prefix`. 

We need this as internally, we use the file name of exported profiles as last-resort metadata source, and the current behaviour of using timezone time is not machine friendly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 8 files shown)


manuelfelipe:
> but I'd appreciate a review from someone who is more of a ruby/rails-ist than I am.

Same for me.  

Added @gmcgibbon for review. Would you mind taking a look to this when you have some time 🙏 ?

gmcgibbon (approved):

manuelfelipe:
^ force-push due to rebase and conflict resolution in changelog. 

added changes from @bmansoob in #95

Issue #96, opened 2023-10-18
[Server] Flaky test 


bmansoob:
I was looking at the flaky test here https://github.com/Shopify/app_profiler/actions/runs/6561028589/attempts/1 and was able to see (via printf debuging) that the finaliser is getting called mid test. https://github.com/Shopify/app_profiler/blob/329d2922298467b77e302b22b951bb02fddabd65/lib/app_profiler/server.rb#L220.

I was able to reliably reproduce [this](https://github.com/Shopify/app_profiler/compare/flaky-test-reproduce?expand=1) by forcing a GC run. Looking at this https://ruby-doc.org/core-3.0.1/ObjectSpace.html#method-c-define_finalizer though, the finaliser should not run on GC?

>Note that if your finalizer references the object to be finalized it will never be run on [GC](https://ruby-doc.org/core-3.0.1/GC.html), although it will still be run at exit. You will get a warning if you capture the object to be finalized as the receiver of the finalizer.

Thoughts @byroot?

casperisfine:
> the finaliser should not run on GC?

Why not? If the object on which the finalizer was set is garbage collected, the finalizer will be called during the sweeping phase (at the end of GC).

casperisfine:
I think https://github.com/Shopify/app_profiler/pull/97 would work.

PR #97, opened 2023-10-18
Undefine the Server finalizer when closing

casperisfine:
Fix: https://github.com/Shopify/app_profiler/issues/96

Otherwise the finalizer might trigger when another server is active and using that path.

cc @bmansoob some test don't pass on my machine on `main`, but could you try this? I think that would fix it.

### DIFF
--- a/lib/app_profiler/server.rb
+++ b/lib/app_profiler/server.rb
@@ -225,8 +225,10 @@ module AppProfiler
         end
 
         def stop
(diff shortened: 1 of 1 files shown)


bmansoob:
A few tests failing -- have not yet looked at the reasons it yet.

casperisfine:
> have not yet looked at the reasons yet.

It's easy, it's because I'm an idiot. `stop` must return `nil` apparently.

PR #101, opened 2023-10-19
Make profiler backend modular, add support for vernier profiler

dalehamel:
This aims to add support for the new [vernier](https://github.com/jhawthorn/vernier) ruby profiler.

To facilitate this, It was necessary to make two fairly major structural changes:

- The "Profiler" now has two possible backends, "Stackprof" (the default) and "Vernier
  - This can be overridden by setting AppProfiler.backend
  - The vernier backend is not yet fully featured

```mermaid
     graph TD;
    Profiler-->Backend-->StackprofBackend;
    Profiler-->Backend-->VernierBackend;
```

- Rather than generic instances of `Profile`, `Profile` is now an abstract / mostly virtual base class, with `StackprofProfile` child class replacing the previous functionality of `Profile`. A new `VernierProfile` is also a child class.
  - This was necessary as Vernier returns a `Result` object, not a hash.

```mermaid
graph TD;
    Profile-->StackprofProfile;
    Profile-->VernierProfile;
```

The size of the change appears larger than it is because some files were moved during the refactoring, but changed enough that they appear as new files - the stackprof related code and tests should be more or less unchanged.

# New behaviour

Stackprof remains the default backend. On older rubies that don't support vernier, vernier will not be possible to use.

On ruby 3.2.1+, it will be possible to use the vernier backend by:

- Separately adding vernier to the application gem file
  - NB It is not a gem dependency of app_profiler due to the ruby version constraints
- Explicitly setting the backend to vernier with `Stackprof.backend = AppProfiler::VernierBackend` OR specifying `backend=vernier` via the middleware (either via query param, or header).
- Specifying the "with_backend" param when calling `AppProfiler.run`, eg: `AppProfiler.run(with_backend: AppProfiler::VernierBackend)`

Note that vernier does not support CPU or object mode profiles, but does support wall and a new type of profile, "retained" memory.

It is not possible to concurrently use vernier and stackprof, as this is unsafe and will result in segfaults. A mutex is introduced to `AppProfiler.run` to prevent this.

# 🎩 

To tophat this, it is probably best to use https://github.com/Shopify/app_profiler/pull/111  which provides a means of viewing the new vernier profiles.

Otherwise on its own this provides a means to collect vernier profiles but just not to view them.


lib/app_profiler/profiler.rb
```diff
@@ -1,86 +1,34 @@
...
-
-        return unless started
-
-        stop
-        results
-      ensure
-        # Only stop the profiler if profiling was started in this context.
-        stop if started
+      def run(params = {}, &block)
+        AppProfiler.profiler_backend.run(params, &block)
```
dalehamel, line 7: I assume there has got to be a better ruby-ish way of doing this than just wrapping the call, but I couldn't pick between delegates / forwardable, or trying to do some sort of mix-in thing. For now I figured i'd just do the simplest thing and see what comes out of the review process.

.github/workflows/ci.yml
```diff
@@ -9,9 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - "2.7"
```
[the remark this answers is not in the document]
casperisfine: The solution is to not have vernier as an explictly dependency, but to try to require it and rescue `LoadError`, https://github.com/rails/rails/blob/4cc1de7910dee1dbb0f4abe509a7ca8995ea83ea/activesupport/lib/active_support/cache/redis_cache_store.rb#L3-L10

lib/app_profiler/profiler.rb
bmansoob: I think we can just get rid of this class/file entirely. 

We will need to change methods in `AppProfiler`.

```
   def run(*args, &block)
      Profiler.run(*args, &block)
   end
```

to 

```
   def run(*args, &block)
      profiler_backend.run(*args, &block)
   end
```

.github/workflows/ci.yml
```diff
@@ -20,7 +21,6 @@ jobs:
         ruby-version: ${{ matrix.ruby }}
     - name: Build
       run: |
-        gem install bundler
```
dalehamel, line 23: Was causing failures on ruby 2.7. Doesn't seem to be needed anyways, as the prior step installs bundler.

Gemfile
```diff
@@ -6,6 +6,7 @@ gemspec
 # Specify the same dependency sources as the application Gemfile
 gem("activesupport", "~> 5.2")
 gem("railties", "~> 5.2")
+gem("vernier", "~> 0.3.1") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
```
dalehamel, line 9: kind of ugly, but lets us support both ruby versions prior to vernier support, and post.

lib/app_profiler/profile.rb
```diff
@@ -6,33 +6,35 @@ class Profile
...
       options = INTERNAL_METADATA_KEYS.map { |key| [key, data[:metadata]&.delete(key)] }.to_h
 
-      new(data, **options).tap do |profile|
+      StackprofProfile.new(data, **options).tap do |profile|
         raise ArgumentError, "invalid profile data" unless profile.valid?
       end
     end
 
-    # `data` is assumed to be a Hash.
+    def self.from_vernier(data)
```
dalehamel, line 20: TODO clean up this file, maybe by separating into different files?

Gemfile
```diff
@@ -6,6 +6,7 @@ gemspec
 # Specify the same dependency sources as the application Gemfile
 gem("activesupport", "~> 5.2")
 gem("railties", "~> 5.2")
+gem("vernier", "~> 0.4.0") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
```
dalehamel, line 9: FIXME

lib/app_profiler.rb
```diff
@@ -60,17 +63,75 @@ module Viewer
   mattr_reader :after_process_queue, default: nil
 
   class << self
-    def run(*args, &block)
-      Profiler.run(*args, &block)
+    def run(*args, with_backend: nil, **kwargs, &block)
```
dalehamel, line 66: with_backend optional keyword argument allows for the backend to be changed from the default for just one profiling session.

lib/app_profiler.rb
```diff
@@ -60,17 +63,75 @@ module Viewer
...
+      yield unless acquire_run_lock
```
dalehamel, line 67: Running stackprof and vernier concurrently can cause  a segmentation fault, which is why this mutex is used to be certain that we cannot have a race condition leading to both profilers running at the same time. Previously, we relied on stackprof itself to prevent multiple concurrent sessions.

.github/workflows/ci.yml
```diff
@@ -12,6 +12,7 @@ jobs:
           - "2.7"
           - "3.0"
           - "3.1"
+          - "3.2.1"
```
gmcgibbon, line 15:
```suggestion
          - "3.2"
```
Actions should automatically use the latest 3.2 if we omit the minor.

Gemfile
gmcgibbon, line 9: Usually, gems have multiple gemfiles to test against [like this](https://github.com/Shopify/identity_cache/tree/80741177e6e036424308a97ab11188885ac7aeeb/gemfiles). We can use a 2.7 gemfile and assume the root gemfile will be using the latest ruby.

README.md
```diff
@@ -76,12 +76,14 @@ Rails.application.config.app_profiler.profile_header = "X-Profile"
...
 | Key | Value | Notes |
 | --- | ----- | ----- |
-| profile/mode | Supported profiling modes: `cpu`, `wall`, `object`. | Use `profile` in (1), and `mode` in (2). |
+| profile/mode | Supported profiling modes: `cpu`, `wall`, `object` for stackprof. | Use `profile` in (1), and `mode` in (2). Vernier backend only supports `wall` and `retained` at present time|
 | async | Upload profile in a background thread. When this is set, profile redirect headers are not present in the response.
 | interval | Sampling interval in microseconds. | |
 | ignore_gc | Ignore garbage collection frames | |
 | autoredirect | Redirect request automatically to Speedscope's page after profiling. | |
 | context | Directory within the specified bucket in the selected storage where raw profile data should be written. | Only supported in (2). Defaults to `Rails.env` if not specified. |
+| backend | Profiler to use, either `stackprof` or `vernier`. Defaults to stackprof. Vernier is only available when added separately in the application Gemfile, and the Ruby version is at least 3.2.1 |
```
gmcgibbon, line 85:
```suggestion
| backend | Profiler to use, either `stackprof` or `vernier`. Defaults to `stackprof`. Note that Vernier requires Ruby 3.2.1+ |
```
Do we expect apps to use multiple backends? If not we could just use the accessor configs to be one or the other.

README.md
```diff
@@ -312,6 +314,20 @@ Note that in `development` and `test` modes the file isn't uploaded. Instead, it
 Rails.application.config.app_profiler.middleware_action = AppProfiler::Middleware::UploadAction
 ```
 
+## Profiler backends
+
+It is possible to configure app_profiler to use the [`vernier`](https://github.com/jhawthorn/vernier) or stackprof. To use vernier, it must be added separately in the application Gemfile.
```
gmcgibbon, line 319:
```suggestion
It is possible to configure AppProfiler to use the [`vernier`](https://github.com/jhawthorn/vernier) or [`stackprof`](https://github.com/tmm1/stackprof). To use `vernier`, it must be added separately in the application Gemfile.
```

I wonder if we should be making stackprof a peer dependency as well (not in the gemspec, but required if defined in the bundle like vernier is). Maybe we can do that in a followup.

lib/app_profiler/backend.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def running?
+      raise NotImplementedError
+    end
+  end
+
+  autoload :StackprofBackend, "app_profiler/backend/stackprof"
+  begin
+    require "app_profiler/backend/vernier"
```
gmcgibbon, line 28: Typically, we define peer deps by using `gem` like you've done and make the file that needs to autoloaded. We can't expect that every developer will want to use vernier, so raising a warning here doesn't make sense. Let's remove it and just autoload the constant instead. The `gem` line you have in the vernier backend will raise an appropriate error.

lib/app_profiler/backend/vernier.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+begin
+  gem("vernier", ">= 0.4.0")
+  require "vernier"
+rescue LoadError
```
gmcgibbon, line 6: I don't think this is necessary. Rails doesn't add extra warnings when it uses this pattern and I don't think we need to either.

lib/app_profiler.rb
gmcgibbon:
```suggestion
    def run(*args, backend: nil, **kwargs, &block)
```

dalehamel:
Notes from pairing:

- See if we can push the run lock into the the concrete backends' .start (eg, vernier.start, stackprof.start), and release in stop. Currently only AppProfiler.run is protected
- Move vernier backend to using autoloading and not require

README.md
```diff
@@ -312,6 +314,20 @@ Note that in `development` and `test` modes the file isn't uploaded. Instead, it
...
 ```
 
+## Profiler backends
+
+It is possible to configure AppProfiler to use the [`vernier`](https://github.com/jhawthorn/vernier) or [`stackprof`](https://github.com/tmm1/stackprof). To use `vernier`, it must be added separately in the application Gemfile.
+
+The backend can be selected dynamically at runtime using the `backend` parameter. The default backend to use when this parameter is not specified can be configured with:
+
+```ruby
+AppProfiler.profiler_backend = AppProfiler::StackprofBackend # or AppProfiler::VernierBackend
```
dalehamel, line 324: TODO update to backend=

lib/app_profiler/backend/stackprof.rb
```diff
@@ -0,0 +1,93 @@
...
+    ensure
+      # Only stop the profiler if profiling was started in this context.
+      stop if started
+    end
+
+    def start(params = {})
+      # Do not start the profiler if StackProf was started somewhere else.
+      return false if running?
+
+      acquire_run_lock
```
dalehamel, line 37: FIXME - return false if we cannot acquire the lock

lib/app_profiler/profile.rb
```diff
@@ -79,11 +96,14 @@ def path
...
+      ].compact.join("-") << format
 
       raise UnsafeFilename if /[^0-9A-Za-z.\-\_]/.match?(filename)
 
       AppProfiler.profile_root.join(filename)
     end
   end
+
+  Profile = AbstractProfile
+  deprecate_constant :Profile
```
dalehamel, line 108: @gmcgibbon I couldn't get the [rails one](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation/DeprecatedConstantAccessor.html#method-i-deprecate_constant) to work as expected, when I tried to test it out it would give the deprecation warning, but then also cause an error. I'm just using the basic ruby one instead as I couldn't figure out why the other one kept exploding.

dalehamel:
@gmcgibbon I believe I've addressed all the comments from our pairing session, please take another look when you have a chance ❤️

README.md
```diff
@@ -76,12 +76,14 @@ Rails.application.config.app_profiler.profile_header = "X-Profile"
 
 | Key | Value | Notes |
 | --- | ----- | ----- |
-| profile/mode | Supported profiling modes: `cpu`, `wall`, `object`. | Use `profile` in (1), and `mode` in (2). |
+| profile/mode | Supported profiling modes: `cpu`, `wall`, `object` for stackprof. | Use `profile` in (1), and `mode` in (2). Vernier backend only supports `wall` and `retained` at present time|
```
gmcgibbon, line 79:
```suggestion
| profile/mode | Supported profiling modes: `cpu`, `wall`, `object` for stackprof. | Use `profile` in (1), and `mode` in (2). Vernier backend only supports `wall` and `retained` at present time. |
```

README.md
```diff
@@ -76,12 +76,14 @@ Rails.application.config.app_profiler.profile_header = "X-Profile"
...
 | async | Upload profile in a background thread. When this is set, profile redirect headers are not present in the response.
 | interval | Sampling interval in microseconds. | |
 | ignore_gc | Ignore garbage collection frames | |
 | autoredirect | Redirect request automatically to Speedscope's page after profiling. | |
 | context | Directory within the specified bucket in the selected storage where raw profile data should be written. | Only supported in (2). Defaults to `Rails.env` if not specified. |
+| backend | Profiler to use, either `stackprof` or `vernier`. Defaults to `stackprof`. Note that Vernier requires Ruby 3.2.1+ |
```
gmcgibbon, line 85:
```suggestion
| backend | Profiler to use, either `stackprof` or `vernier`. Defaults to `stackprof`. Note that Vernier requires Ruby 3.2.1+. |
```

lib/app_profiler/backend.rb
```diff
@@ -0,0 +1,47 @@
...
+    def release_run_lock
+      self.class.run_lock.unlock
+    rescue ThreadError
+      AppProfiler.logger.warn("[AppProfiler] run lock not released as it was never acquired")
+    end
+  end
+
+  autoload :StackprofBackend, "app_profiler/backend/stackprof"
+  autoload :VernierBackend, "app_profiler/backend/vernier"
+  DefaultBackend = AppProfiler::StackprofBackend
```
gmcgibbon, line 46: Let's nest these inside of `AppProfiler::Backend` instead of defining them in `AppProfiler` inside of the file that's supposed to only define `AppProfiler::Backend`. We aren't using Zeitwerk yet, but I'd like to remain compatible as much as possible (like how viewer and storage constants are defined).

Also, this triggers the autoload immediately, so I don't think it is necessary to define `DefaultBackend`. We can just set `@profiler_backend ||= Backend::Stackprof`, which will be evaluated when the method is called.

lib/app_profiler/backend/stackprof.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require "stackprof"
+
+module AppProfiler
+  class StackprofBackend < Backend
```
gmcgibbon, line 6: The naming here is wrong. Let's rename to `AppProfiler::Backend::Stackprof` to match the file name (also like how viewer and storage are defined).

lib/app_profiler/middleware.rb
```diff
@@ -31,7 +31,7 @@ def profile(env, params)
 
       return yield unless before_profile(env, params_hash)
 
-      profile = AppProfiler.run(params_hash) do
+      profile = AppProfiler.run(params_hash, backend: params.backend) do
```
gmcgibbon, line 34: We should probably be changing the behaviour of `params.to_h` instead of explicitly setting the backend, right?

lib/app_profiler/profile.rb
gmcgibbon: Does that happen in a rails app? Check to make sure the deprecation behaviour in the application is set to log, if you want it to log. I think the fact that it was raising is dependent on the app's config.

lib/app_profiler/profile/stackprof.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module AppProfiler
+  class StackprofProfile < AbstractProfile
+    FILE_EXTENSION = ".stackprof.json"
```
gmcgibbon, line 5: How would this handle legacy profile names without this extension?

test/app_profiler/backend_test.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module AppProfiler
+  class BackendTest < TestCase
+    test ".backend= fails to update the backend if already profiling" do
+      skip("Vernier not supported") unless defined?(AppProfiler::VernierBackend::NAME)
```
gmcgibbon, line 8: Let's wrap the entire test case in a ruby version if statement instead.

lib/app_profiler/request_parameters.rb
```diff
@@ -16,17 +16,29 @@ def async
...
     def valid?
       if mode.blank?
         return false
       end
 
-      unless Parameters::MODES.include?(mode)
-        AppProfiler.logger.info("[Profiler] unsupported profiling mode=#{mode}")
+      return false if backend != AppProfiler::StackprofBackend::NAME && !defined?(AppProfiler::VernierBackend::NAME)
+
+      if defined?(AppProfiler::VernierBackend::NAME) && backend == AppProfiler::VernierBackend::NAME &&
```
gmcgibbon, line 31: Maybe we can make this easier by defining `AppProfiler.vernier_supported?`? I think you can use `autoload?` to check if it is loaded or not.

test/app_profiler/backend_test.rb
dalehamel: A ruby version if statement is not sufficient on its own, we need to verify if vernier is also supported. I updated it to use the new `vernier_supported?` helper though. This should implicitly ensure that the ruby version is also adequate.

lib/app_profiler/profile.rb
dalehamel: Ok to be clear, here is the error I'm getting if I try and use the old constant:

```
DEPRECATION WARNING: AppProfiler::Profile is deprecated! Use AbstractProfile instead. (called from <top (required)> at /opt/rubies/3.2.1/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11)
/Users/dalehamel/.gem/ruby/3.2.1/gems/activesupport-5.2.4.3/lib/active_support/inflector/methods.rb:283:in `const_get': uninitialized constant AbstractProfile (NameError)

          constant.const_get(name)
                  ^^^^^^^^^^
```

So it gives a warning, then it shits the bed trying to find the constant.

Here's how I am using it:

```
...

107   include ActiveSupport::Deprecation::DeprecatedConstantAccessor
108   deprecate_constant 'Profile', 'AbstractProfile', deprecator: ActiveSupport::Deprecation.new

```

It is fine to just use `AbstractProfile` directly.

dalehamel:
> Does that happen in a rails app? Check to make sure the deprecation behaviour in the application is set to log, if you want it to log. I think the fact that it was raising is dependent on the app's config.

Added more details in https://github.com/Shopify/app_profiler/pull/101#discussion_r1496535200

Thanks for your great review @gmcgibbon , sorry it took me a week to get back to it but I was sick all of last week. I believe I've addressed all your comments either with code or additional clarifying comments, and I added some clarity around the deprecation issues I am having.

Please take a look when you have another chance - thanks!

lib/app_profiler/middleware.rb
gmcgibbon: I'm not sure I understand. Can't you just use `**params_hash` to have the hash satisfy the kwarg? You shouldn't need to change the signature:
```
irb(main):001* def run(*args, backend: nil, **kwargs)
irb(main):002*   puts args.inspect, backend, kwargs.inspect
irb(main):003> end
=> :run
irb(main):004> run(**{ backend: "hello", other: :opt })
[]
hello
{:other=>:opt}
=> nil
```

lib/app_profiler.rb
```diff
@@ -29,11 +32,16 @@ module Viewer
     autoload :SpeedscopeRemoteViewer, "app_profiler/viewer/speedscope_remote_viewer"
   end
 
+  module Backend
+    autoload :Stackprof, "app_profiler/backend/stackprof"
```
gmcgibbon, line 36:
```suggestion
    autoload :StackprofBackend, "app_profiler/backend/stackprof_backend"
```
I think this will help with not having to qualify `::Stackprof` inside this module, and also disambiguating from the stackprof constant. 

We should follow the same pattern for vernier. Your test files already seem to be named like this.

lib/app_profiler/backend/vernier.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+gem("vernier", ">= 0.4.0")
+require "vernier"
+
+module AppProfiler
+  module Backend
+    class Vernier < Base
+      NAME = "vernier"
```
gmcgibbon, line 9: So personally, I think we should be using symbols here. It will allocate less memory, and it is easier to type when specifying for configuration. Usually, config values are symbols over strings too. When assigning the backend, you can just cast via to_sym.

lib/app_profiler/profile/stackprof.rb
gmcgibbon: Right. Sounds good, thanks! I'm going to focus on reviewing this one first. Perhaps we can punt this to the other PR if it only needed for that?

lib/app_profiler.rb
```diff
@@ -60,17 +68,77 @@ module Viewer
...
   class << self
-    def run(*args, &block)
-      Profiler.run(*args, &block)
+    def run(*args, backend: nil, **kwargs, &block)
+      orig_backend = self.backend
+      begin
+        self.backend = backend if backend
+        profiler.run(*args, **kwargs, &block)
+      rescue BackendError
+        yield
```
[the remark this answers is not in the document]
dalehamel: The premise behind that is that we don't want profiling to cause a request that would otherwise succeed to fail.

Maybe to avoid obliquely swallowing the error, we can just log that this occurred here.

lib/app_profiler/middleware.rb
dalehamel: yes so as you've defined it there it works fine as when you do the splat, the backend parameter is extracted into a variable that you can use for the intended purpose of toggling the backend, and the other parameters are forwarded to the actual profiler.

I had assumed by your earlier comment:

> We should probably be changing the behaviour of params.to_h instead of explicitly setting the backend, right?

That you thought we should drop that change to the **signature** too. But i guess you just meant how we generate the hash that it is called with.

That is a much simpler change to make, my apologies for the misunderstanding - i'll make the change.

Gemfile
dalehamel: Note to self - bump this to 0.5.0 now that it has been released

lib/app_profiler.rb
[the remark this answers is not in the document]
gmcgibbon: Hm, I wonder what happens when you supply bad values to things like incorrect modes today. Do we need to rescue other error types?

lib/app_profiler.rb
```diff
@@ -60,17 +68,77 @@ module Viewer
...
+    end
+
+    def profiler
+      @backend ||= backend.new
+    end
+
+    def backend=(new_backend)
+      new_profiler_backend = if new_backend.is_a?(String)
+        backend_for(new_backend)
+      elsif new_backend&.< Backend::Base
```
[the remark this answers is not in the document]
dalehamel: Yeah let's go with just accepting the symbol.

dalehamel:
@gmcgibbon I believe I addressed your most recent batch of comments, please take another look when you have a chance. Sorry for the long delay on the back and forth again, I had some vacation that prevented as quick of a turn-around as I would have liked.

lib/app_profiler/railtie.rb
```diff
@@ -40,6 +40,7 @@ class Railtie < Rails::Railtie
       AppProfiler.profile_enqueue_success = app.config.app_profiler.profile_enqueue_success
       AppProfiler.profile_enqueue_failure = app.config.app_profiler.profile_enqueue_failure
       AppProfiler.after_process_queue = app.config.app_profiler.after_process_queue
+      AppProfiler.backend = app.config.app_profiler.profiler_backend || :stackprof
```
gmcgibbon, line 43:
```suggestion
      unless AppProfiler.running?
        AppProfiler.backend = app.config.app_profiler.profiler_backend || :stackprof
      end
```
For boot time profiles, otherwise it crashes.

dalehamel: @gmcgibbon why does it crash? maybe we need to also set this constant in lib/app_profiler.rb?

dalehamel: The first time we reference the backend, if it is nil, it should default to stackprof already:

https://github.com/Shopify/app_profiler/pull/101/files#diff-bbd7626aa479b35f000a50b4e1c7d1f660c1eaaf1fa2c2ed71a1f12775a239afR130

So i'd like to see the backtrace for the boot time crash to ensure safety if this is not adequate

dalehamel: or, is the crash happening because it tries to update the backend while profiling is already happening? If so, i'd like to make that impossible rather than just guard for it here.

lib/app_profiler/profile.rb
```diff
@@ -79,11 +96,14 @@ def path
...
-      ].compact.join("-") << ".json"
+      ].compact.join("-") << format
 
       raise UnsafeFilename if /[^0-9A-Za-z.\-\_]/.match?(filename)
 
       AppProfiler.profile_root.join(filename)
     end
   end
+
+  include ActiveSupport::Deprecation::DeprecatedConstantAccessor
```
gmcgibbon, line 107: We need a require for this.

lib/app_profiler.rb
```diff
@@ -29,11 +32,16 @@ module Viewer
     autoload :SpeedscopeRemoteViewer, "app_profiler/viewer/speedscope_remote_viewer"
   end
 
+  module Backend
+    autoload :StackprofBackend, "app_profiler/backend/stackprof_backend"
+    autoload :VernierBackend, "app_profiler/backend/vernier_backend"
+  end
```
gmcgibbon, line 38: If we're requiring backend below, let's put the autoloads in that file instead.

lib/app_profiler.rb
```diff
@@ -60,17 +68,78 @@ module Viewer
...
+      @profiler_backend ||= Backend::StackprofBackend
+      @profiler_backend::NAME
+    end
+
+    def vernier_supported?
+      defined?(AppProfiler::Backend::VernierBackend::NAME)
+    end
+
+    def clear
+      @backend.stop if @backend&.running?
```
gmcgibbon, line 141: I think this might be overly defensive. If we're trying to assign a new backend and the backend is running, shouldn't we raise instead of stopping?

lib/app_profiler/backend.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module AppProfiler
+  module Backend
+    class Base
```
gmcgibbon, line 5: Let's move this to `lib/app_priofiler/backend/base.rb` and put autoloads for `Base`, `StackprofBackend`, `VernierBackend`. Also nitpick that the naming should either be `BaseBackend` and `<Name>Backend` as the subclass, or `Base` and `<Name>` as the subclass.

lib/app_profiler/backend/stackprof_backend.rb
```diff
@@ -3,13 +3,21 @@
...
 
 module AppProfiler
-  module Profiler
-    DEFAULTS = {
-      mode: :cpu,
-      raw: true,
-    }.freeze
+  module Backend
+    class StackprofBackend < Base
+      NAME = :stackprof
```
gmcgibbon, line 8: I think this should be a method. If we ever use Sorbet in this project, it would be easier to enforce subclasses implement this by exposing a method.

lib/app_profiler/profile.rb
```diff
@@ -1,42 +1,45 @@
 # frozen_string_literal: true
 
+require "active_support/deprecation/constant_accessor"
+
 module AppProfiler
-  class Profile
+  autoload :StackprofProfile, "app_profiler/profile/stackprof"
+  autoload :VernierProfile, "app_profiler/profile/vernier"
+
+  class AbstractProfile
```
gmcgibbon, line 9: It would be better to name this `BaseProfile` like we have with backends.

lib/app_profiler/backend/stackprof_backend.rb
```diff
@@ -35,31 +44,38 @@ def start(params = {})
...
       end
 
       def results
-        stackprof_profile = stackprof_results
+        stackprof_profile = backend_results
 
         return unless stackprof_profile
 
-        Profile.from_stackprof(stackprof_profile)
+        AppProfiler::AbstractProfile.from_stackprof(stackprof_profile)
```
gmcgibbon, line 64:
```suggestion
        AbstractProfile.from_stackprof(stackprof_profile)
```
🤔 I'm not a huge fan of calling methods from an abstract class. I wonder if we're better off keeping the name `Profile`, but we can fix this in another PR.

gmcgibbon (commented):
LGTM after some name changes and file moves.

lib/app_profiler/backend/stackprof_backend.rb
dalehamel: @gmcgibbon this is the reason the deprecator was introduced, so if we go back to just "Profile" we don't need the deprecator stuff anymore.

A couple of ways to do this I can think of:

- Just create a subclass of "BaseProfile" called "Profile"
- Alias "BaseProfile" to "Profile"
- Make BaseProfile a module, include that in "Profile", and make "Profile" the base class
- Instead of calling it BaseProfile, just call it Profile

I don't really care or have a preference as to which, what do you think is the most idiomatic?

I agree that both "AbstractProfile.from..." and "BaseProfile.from..." look weird though.

lib/app_profiler/backend/stackprof_backend.rb
dalehamel: Moved it to a class method

lib/app_profiler.rb
dalehamel: Err wait, i remember why i did this now - it's to prevent a running backend from becoming orphaned via the clear method. If we lose a reference to the backend, and a session is running, we can NEVER end the session. This is to ensure that doesn't happen.

dalehamel:
@gmcgibbon I think i've done the renames and moves as you requested, and I replied to your two other suggestions.

The CI bundle cache from main seems to have borked somehow, which is why everything but ruby 3.2 (introduced here) is failing. Not sure how best to fix that, but that is why there is red in the tests. I guess I'll take a look at that next to try and get this to green 😓

dalehamel:
Ah the failures seem legit, I guess we need something like https://github.com/Shopify/app_profiler/pull/101#discussion_r1471901897 - one for the older rubies, one for the newer ones...

dalehamel:
Ok I've got it back to green by:

- Tricking older rubies into using a separate lockfile in CI using `BUNDLE_GEMFILE` and the test matrix
- Explicitly splatting an empty params hash in a test, or else ruby 2.7 is sad as the expectation isn't satisfied

lib/app_profiler/backend/stackprof_backend.rb
gmcgibbon: Let's rename to `BaseProfile` so that it is more aligned with how the rest of the gem is structured. You can call the from... methods in the subclasses or the base class, which I think it fine. I just don't like the "Abstract" naming here because it doesn't sound like Ruby to me.

test/app_profiler/profile/stackprof_test.rb
```diff
@@ -23,75 +23,74 @@ class ProfileTest < TestCase
...
       assert_equal("object", profile.mode)
     end
 
     test "#view" do
-      profile = Profile.new(stackprof_profile)
+      profile = StackprofProfile.new(stackprof_profile)
 
-      AppProfiler.stubs(:viewer).returns(Viewer::SpeedscopeViewer)
-      Viewer::SpeedscopeViewer.expects(:view).with(profile)
+      Viewer::SpeedscopeViewer.expects(:view).with(profile, **{})
```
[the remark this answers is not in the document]
gmcgibbon: Actually, on second thought since our min ruby is 2.7 and we have no real reason to drop 2.7 for the lib code yet we can keep it for now.

Gemfile-legacy
```diff
@@ -0,0 +1 @@
+Gemfile
```
gmcgibbon, line 1: Instead of symlinking, let''s get rid of the conditional logic in the main gemfile and just remove vernier from other legacy one. It ismore scalable as we add more gemfiles to the CI matrix. We should really be testing different Rails versions.

test/app_profiler/profile/stackprof_test.rb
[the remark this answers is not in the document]
dalehamel: Ok i'll just leave this as is then.

Gemfile-legacy
dalehamel: ok

lib/app_profiler.rb
dalehamel: Made it private, and added a test.

lib/app_profiler/backend/base.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module AppProfiler
+  module Backend
+    autoload :StackprofBackend, "app_profiler/backend/stackprof_backend"
+    autoload :VernierBackend, "app_profiler/backend/vernier_backend"
```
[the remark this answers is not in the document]
gmcgibbon: Yes, you fix it. Thanks!

PR #111, opened 2024-01-24
Add firefox viewer, simplify speedscope viewer options

dalehamel:
Depends on https://github.com/Shopify/app_profiler/pull/101


lib/app_profiler/railtie.rb
```diff
@@ -41,13 +40,13 @@ class Railtie < Rails::Railtie
...
       AppProfiler.backend = app.config.app_profiler.profiler_backend || AppProfiler::DefaultBackend
+      AppProfiler.gecko_viewer_package = app.config.app_profiler.gecko_viewer_package || "https://github.com/firefox-devtools/profiler"
     end
 
     initializer "app_profiler.add_middleware" do |app|
       unless AppProfiler.middleware.disabled
-        if AppProfiler.viewer == Viewer::SpeedscopeRemoteViewer
-          app.middleware.insert_before(0, Viewer::SpeedscopeRemoteViewer::Middleware)
-        end
+        app.middleware.insert_before(0, Viewer::SpeedscopeViewer::Middleware)
```
dalehamel, line 48: TODO ensure that these are not added in production

lib/app_profiler.rb
```diff
@@ -28,8 +28,8 @@ module Storage
 
   module Viewer
     autoload :BaseViewer, "app_profiler/viewer/base_viewer"
-    autoload :SpeedscopeViewer, "app_profiler/viewer/speedscope_viewer"
-    autoload :SpeedscopeRemoteViewer, "app_profiler/viewer/speedscope_remote_viewer"
```
gmcgibbon, line 32: If I'm reading this right, we are removing the remote viewer. It is public API, so let's revert that.

lib/app_profiler/viewer/base_viewer.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+gem "rails-html-sanitizer", ">= 1.6.0"
+require "rails-html-sanitizer"
```
gmcgibbon, line 4: This would require folks using the basic speedscope viewer to require rails-html-sanitizer, but it isn't actually needed. I think this is the wrong place for this. We can make a subclass for remote profile viewing instead.

lib/app_profiler/yarn/command.rb
```diff
@@ -5,15 +5,20 @@ module Yarn
     module Command
       class YarnError < StandardError; end
 
+      GECKO_VIEWER_PACKAGE = AppProfiler.gecko_viewer_package
```
gmcgibbon, line 8: Assigning a configurable value to a constant doesn't really make sense to me. We should just inline the string here instead.

lib/app_profiler.rb
dalehamel: @gmcgibbon as the PR description says, the two are consolidated. What was formerly the "remote" viewer is now the only option as I don't see any benefit to having two ways to do the same thing.

I understand it's a public API change, i think releasing as 0.2.0 with a note would make sense.

Likewise, it also removes the viewer being configurable (AppProfiler.viewer config will just be ignored) - stackprof will use speedscope, vernier will use firefox-profiler.

dalehamel:
notes from pairing:

- Bring back the local viewer
- Rename current speedscope viewer back to remote, corresponding name change for firefox-profiler also
- Documentation notice about `/from-url` path being taken for firefox-profiler

lib/app_profiler/yarn/command.rb
```diff
@@ -29,6 +31,14 @@ def setup_yarn
         yarn("init", "--yes") unless package_json_exists?
       end
 
+      def yarn_setup
+        @yarn_initialized || false
+      end
+
+      def yarn_setup=(state)
```
dalehamel, line 38: only used in testing to fake that we are already initialized.

lib/app_profiler/yarn/command.rb
```diff
@@ -29,6 +31,14 @@ def setup_yarn
         yarn("init", "--yes") unless package_json_exists?
       end
 
+      def yarn_setup
```
dalehamel, line 34: this is now instance level, not a module singleton's variable.

lib/app_profiler/profile/vernier.rb
```diff
@@ -17,7 +17,7 @@ def format
     end
 
     def view(params = {})
-      raise NotImplementedError
+      Viewer::FirefoxRemoteViewer.view(self, **params)
```
gmcgibbon, line 20: `AppProfiler.vernier_viewer.view(self, **params)`

gmcgibbon: The developer should not be forced into a single viewer. Ideally, we should have a non-remote viewer that is default that opens in a web browser like the regular speedscope viewer does.

lib/app_profiler/viewer/base_viewer.rb
gmcgibbon: This is currently not in the right place. We should only make those interested in remote viewers have this as a dependency. Placing it in base viewer will require it for everyone, right?

lib/app_profiler/viewer/middleware/firefox.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "app_profiler/yarn/command"
+require "app_profiler/yarn/with_firefox_profiler"
+
+module AppProfiler
+  module Viewer
+    class FirefoxRemoteViewer < BaseViewer
```
gmcgibbon, line 8:
```suggestion
    class VernierRemoteViewer < BaseViewer
```

dalehamel: The actual viewer is firefox, not vernier. We don't call speedscope "stackprof viewer"

lib/app_profiler/viewer/base_viewer.rb
dalehamel: Moved this in ad67c69 to not be part of the base viewer.

lib/app_profiler/viewer/middleware/firefox.rb
gmcgibbon: Yes, you're right. It makes it confusing now that there's multiple backends and viewers. We should probably add docs to each viewer indicating which backend it is meant for.

lib/app_profiler.rb
```diff
@@ -53,8 +55,10 @@ module Viewer
   mattr_reader   :profile_url_formatter,
     default: DefaultProfileFormatter
 
+  mattr_accessor :gecko_viewer_package, default: "https://github.com/tenderlove/profiler#v0.0.2"
   mattr_accessor :storage, default: Storage::FileStorage
-  mattr_accessor :viewer, default: Viewer::SpeedscopeViewer
+  mattr_accessor :viewer, default: Viewer::SpeedscopeViewer # DEPRECATED
+  mattr_accessor :speedscope_viewer, default: Viewer::SpeedscopeViewer
```
gmcgibbon, line 61:
```suggestion
  mattr_accessor :stackprof_viewer, default: Viewer::SpeedscopeViewer
```
This distinction should be per backend, not per viewer type

lib/app_profiler.rb
```diff
@@ -179,6 +182,17 @@ def profile_url(upload)
 
       AppProfiler.profile_url_formatter.call(upload)
     end
+
+    # DEPRECATIONS
+    def viewer
+      ActiveSupport::Deprecation.warn("viewer is deprecated, use speedscope_viewer instead")
```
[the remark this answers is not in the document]
gmcgibbon: https://github.com/Shopify/app_profiler/pull/111/files#r1813934468

lib/app_profiler/profile/vernier.rb
gmcgibbon: https://github.com/Shopify/app_profiler/tree/firefox_viewer_local

dalehamel: @gmcgibbon i think i misunderstand what you mean by remote / local.

There is no way to just "yarn run" firefox-profiler with a profile like you can with speedscope, it needs a web server. But we already have a web server from the app we are profiling in most cases. If by "local" you mean it includes its own web server, lets do that in another PR - this one is complex enough and has been open for long enough, and the current implementation will satisfy most users.

gmcgibbon: Yes, I'm open to doing this in another PR but I think we still need to do this work is what I'm saying in order to close the issue I opened. The DX of the remote middleware isn't nearly as good.

gmcgibbon: Let's apply some of the other fixes I have here though: https://github.com/Shopify/app_profiler/compare/firefox-viewer...firefox_viewer_local

It introduces a proper deprecator, and fixes some bugs in this approach.

dalehamel: > Let's apply some of the other fixes I have here though: [firefox-viewer...firefox_viewer_local](https://github.com/Shopify/app_profiler/compare/firefox-viewer...firefox_viewer_local)
> 
> It introduces a proper deprecator, and fixes some bugs in this approach.

You seem to be moving much faster than me on this, wanna just push directly to this branch? Will be faster than telling me what needs to be done, I'm not a ruby or rails dev.

gmcgibbon:
Everything should be fixed aside from minor linting. Though, the remote middleware doesn't seem to work. I'll put this out of draft when it works.

gmcgibbon:
Hm, I'm not the author so I can't draft it. I'll ping again when it is ready for review.

lib/app_profiler/middleware.rb
```diff
@@ -4,7 +4,6 @@
 require "app_profiler/middleware/base_action"
 require "app_profiler/middleware/upload_action"
 require "app_profiler/middleware/view_action"
-require "app_profiler/sampler/config"
```
dalehamel, line 7: I don't think we meant to remove this?

cc @bmansoob

dalehamel (commented):
Looks good except for removing the sampler from the middleware in 148aa55, this is unrelated to the intent of this PR and i think possibly just an accident?

lib/app_profiler/middleware.rb
gmcgibbon: I did, it is autoloaded now because the profile called the deprecated but was required after the deprecator was defined. So I allowed all toplevel constants to autoload (like this one, `AppProfiler::Sampler`).

dalehamel: Ah makes sense, didn't see that it moved to the autoload 👍  just wanted to make sure we didn't inadvertently add a regression.

dalehamel:
Looks good, thanks @gmcgibbon

lib/app_profiler/railtie.rb
```diff
@@ -11,7 +11,12 @@ class Railtie < Rails::Railtie
       AppProfiler.logger = app.config.app_profiler.logger || Rails.logger
       AppProfiler.root = app.config.app_profiler.root || Rails.root
       AppProfiler.storage = app.config.app_profiler.storage || Storage::FileStorage
-      AppProfiler.viewer = app.config.app_profiler.viewer || Viewer::SpeedscopeViewer
+      if app.config.app_profiler.stackprof_viewer
```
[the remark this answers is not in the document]
gmcgibbon: I changed this because the remote firefox viewer has a peer dependency we may or may not need. It depends which backend the developers uses. This way it won't be loaded until it is used.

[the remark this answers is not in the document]
bmansoob: 👍

PR #116, opened 2024-04-16
Let setup-ruby action handle bundler

DazWorrall:
It resolves some CI issues with ruby 2.7 ([e.g.](https://github.com/Shopify/app_profiler/actions/runs/8335697996/job/22811641704?pr=114)) and also sets up caching for faster future builds.


DazWorrall:
Sorry @dalehamel - revert it if you need to, I am afk

PR #124, opened 2024-06-18
Enforce ruby version requirement with Gemfile for dependabot

dalehamel:
It looks like dependabot tasks run with an ancient Ruby, and it isn't happy with the vernier dependency in the Gemfile.

See: https://github.com/Shopify/app_profiler/actions/runs/9566253375

To work around this, vernier will only be included on versions that actually support it.


dalehamel:
We'd hoped to avoid the need for this with separate Gemfiles but I can't figure out how to specify the ruby version for dependabot to use. I'll file an issue to keep track of this.

PR #126, opened 2024-07-25
Forward profile metadata to GCS

bmansoob:
This will help in downstream ingestion processes where there is a need to extract profiles' metadata and do something with it, without having to deserialise the actual profile. 

We have such a use-case; we use values in metadata as filters in Observe.

### DIFF
--- a/test/app_profiler/storage/google_cloud_storage_test.rb
+++ b/test/app_profiler/storage/google_cloud_storage_test.rb
@@ -22,6 +22,16 @@ module AppProfiler
         end
       end
 
+      test "sets metadata when forward_metadata_on_upload" do
+        with_forward_metadata_on_upload do
+          profile = profile_from_stackprof
+          with_mock_gcs_bucket(profile) do
+            uploaded_file = GoogleCloudStorage.upload(profile)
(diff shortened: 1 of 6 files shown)


dalehamel (approved):

test/app_profiler/storage/google_cloud_storage_test.rb
```diff
@@ -22,6 +22,31 @@ def teardown
...
+          AppProfiler.forward_metadata_on_upload = forward_metadata_on_upload
+
+          file = Google::Cloud::Storage::File.new
+          profile = profile_from_stackprof
+          bucket = Google::Cloud::Storage::Bucket.new
+
+          expected_metadata = nil
+          if forward_metadata_on_upload
+            expected_metadata = profile.metadata
+          end
```
gmcgibbon, line 37: This is a bit of a code smell, you're duplicating the logic in your test to save a few lines, and that can make the test brittle over time. I'd consider changing `with_mock_gcs_bucket` to assert properties you need, and introducing a method like `with_stubbed_process_queue_thread` below for your new config to toggle global values.

bmansoob:
commits squashed

PR #128, opened 2024-08-13
add profile sampling

bmansoob:


manuelfelipe (approved):
everything looks great (for my un-trained RoR eyes) based on what we discussed. If i were to nit something it would be on the config names but decided not to in the spirit of moving on.

bmansoob:
thanks, @manuelfelipe. Please feel free to suggest -- should be easy enough to change.

README.md
```diff
@@ -336,6 +336,50 @@ Rails.application.config.app_profiler.backend = AppProfiler::StackprofBackend #
...
+
+Apps can disable profile sampling on the fly (killswitch, beta flag etc) via `profile_sampler_status_proc`:
+
+```ruby
+AppProfiler.profile_sampler_status_proc = -> { ... }
+# OR
+Rails.application.config.app_profiler.profile_sampler_status_proc = -> { ... }
+```
+
+Sampling configuration paramets along with their default values
```
dalehamel, line 360:
```suggestion
Sampling configuration parameters along with their default values:
```

README.md
```diff
@@ -336,6 +336,50 @@ Rails.application.config.app_profiler.backend = AppProfiler::StackprofBackend #
...
+
+| Parameter                           | Default |
```
dalehamel, line 362: What about on-CPU?

lib/app_profiler/sampler/config.rb
```diff
@@ -0,0 +1,134 @@
...
+      ## Default values
+      SAMPLE_RATE = 0.001 # 0.1%
+      PATHS = ["/"]
+
+      WALL_INTERVAL = 5000
+      RETAINED_INTERVAL = 0
+      OBJECT_INTERVAL = 1000
+
+      WALL_MODE_PROBABILITY = 0.8 # 80.0%
+      OBJECT_OR_RETAINED_MODE_PROBABILITY = 0.2 # 20.0%
```
dalehamel, line 80: Not sure I like this, we might see something more analogous to object mode in the future with vernier.

Also, it is possible for wall mode profiles to sample allocations similar to object mode in recent versions.

dalehamel (commented):
Overall looks good, but I'm not sure how we should handle the differences in stackprof and vernier, and the available configurations.

lib/app_profiler/sampler/vernier_config.rb
```diff
@@ -0,0 +1,44 @@
...
+module AppProfiler
+  module Sampler
+    class VernierConfig
+      attr_reader :modes_probability
+
+      WALL_INTERVAL = 5000
+      RETAINED_INTERVAL = 5000
+
+      WALL_MODE_PROBABILITY = 0.8
+      RETAINED_MODE_PROBABILITY = 0.2
```
dalehamel, line 12: this is probably too high, the impact of retained profiling is orders of magnitude more than stackprof object profiling. I think the default for this should be 0, and we can assume unless people asked for retained profiling the want wall mode.

In the future we can also consider a third "mode", which is "allocation_sample_rate" done in wall mode, seems to be some kind of a hybrid of wall and stackprof object profile?

lib/app_profiler.rb
```diff
@@ -62,6 +63,9 @@ module Viewer
   mattr_reader :profile_enqueue_failure, default: nil
   mattr_reader :after_process_queue, default: nil
   mattr_accessor :forward_metadata_on_upload, default: false
+  mattr_accessor :profile_sampler_enabled, default: false
+  mattr_accessor :profile_sampler_status_proc, default: nil
```
bmansoob, line 67: open to suggestion for this - do not like ending it with `proc`, but kind of want to convey what it actually is.

README.md
bmansoob: cpu mode got added after the whole refactor

lib/app_profiler/sampler.rb
```diff
@@ -0,0 +1,60 @@
...
+        get_profile_params(config, random)
+      end
+
+      private
+
+      def sample?(random, config, request)
+        return false if random > config.sample_rate
+
+        path = request.path
+        return false unless config.paths.any? { |p| path.match?(p) }
```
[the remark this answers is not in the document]
bmansoob: Thanks, Jean. Just to confirm the suggestion is change `paths` from an array of strings to an array of Procs? 🤔

lib/app_profiler.rb
bmansoob: I will actually remove it from this version and keep everything in static config. When we decide to port Core over to use the app_profiler sampler, we will need to add functionality to fetch the config dynamically. So, it is better to add something like this then.

lib/app_profiler/sampler.rb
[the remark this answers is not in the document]
casperisfine: No the suggestion is to use `===` to match the regexp, so that it allow users to use something else than regexp if they need/want to.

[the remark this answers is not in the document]
casperisfine: I misunderstood that `config.path` was a lit of regexp, actually it's a list of string, in such context `.match?` is the most flexible. Ignore that part.

PR #130, opened 2024-08-28
Sampler - fix probability logic

bmansoob:
There was a bug in how `Kernel.rand` got passed to every function.

One of the first things we check:

`return false if random > config.sample_rate`

If the `sample_rate` is something like 0.001 then only lower probabilities options (mode, backend) will be chosen (because random was created once and passed down).

Fixed it by calling Kernel.rand every time.


dalehamel (approved):
Makes sense - didn't realize in the prior review we were inadvertently filtering the rands that got passed down

PR #132, opened 2024-09-03
Default to upload action in test environment

casperisfine:
Test environment should pretend to behave like production, not development.


gmcgibbon:
Perhaps we can compromise on making this configurable?

PR #134, opened 2024-09-05
[Sampler] Allow passing in a Proc to enable/disable the sampler.

bmansoob:
Small improvements for the profile sampler:

- `profile_sampler_enabled` accepts a Proc now. The idea is to remove the sampler we have in Core and use this. 
- Renamed `paths` in the config to `targets`. The idea is that the sampler can be use by Jobs too -- so it makes sense to make it generic.

All changes are backwards compatible and should not break existing clients.

### DIFF
--- a/lib/app_profiler.rb
+++ b/lib/app_profiler.rb
@@ -63,7 +63,7 @@ module AppProfiler
   mattr_reader :profile_enqueue_failure, default: nil
   mattr_reader :after_process_queue, default: nil
   mattr_accessor :forward_metadata_on_upload, default: false
-  mattr_accessor :profile_sampler_enabled, default: false
+
   mattr_accessor :profile_sampler_config
 
   class << self
@@ -114,6 +114,28 @@ module AppProfiler
 
       clear
       @profiler_backend = new_profiler_backend
+    end
+
+    def profile_sampler_enabled=(value)
+      if value.is_a?(Proc)
+        raise ArgumentError,
+          "profile_sampler_enabled must be a proc or a lambda that accepts no argument" if value.arity != 0
+      else
+        raise ArgumentError, "Must be TrueClass or FalseClass" unless [TrueClass, FalseClass].include?(value.class)
+      end
+
+      @profile_sampler_enabled = value
+    end
+
+    def profile_sampler_enabled
+      return false unless defined?(@profile_sampler_enabled)
+
+      @profile_sampler_enabled.is_a?(Proc) ? @profile_sampler_enabled.call : @profile_sampler_enabled
+    rescue => e
+      logger.error(
+        "[AppProfiler.profile_sampler_enabled] exception: #{e}, message: #{e.message}",
+      )
+      false
     end
 
     def backend_for(backend_name)
--- a/lib/app_profiler/sampler/config.rb
+++ b/lib/app_profiler/sampler/config.rb
@@ -5,19 +5,20 @@ require "app_profiler/sampler/vernier_config"
 module AppProfiler
   module Sampler
     class Config
-      attr_reader :sample_rate, :paths, :cpu_interval, :backends_probability
+      attr_reader :sample_rate, :targets, :cpu_interval, :backends_probability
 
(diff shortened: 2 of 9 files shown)


lib/app_profiler.rb
```diff
@@ -116,6 +116,25 @@ def backend=(new_backend)
...
+
+      @profile_sampler_enabled = value
+    end
+
+    def profile_sampler_enabled
+      return false unless defined?(@profile_sampler_enabled)
+
+      @profile_sampler_enabled.is_a?(Proc) ? @profile_sampler_enabled.call : @profile_sampler_enabled
+    rescue
+      false
```
bmansoob, line 135: in case Proc blows up

lib/app_profiler/sampler/config.rb
```diff
@@ -5,28 +5,31 @@
...
+        },
+        paths: nil)
 
         if sample_rate < 0.0 || sample_rate > 1.0
           raise ArgumentError, "sample_rate must be between 0 and 1"
         end
 
         raise ArgumentError, "mode probabilities must sum to 1" unless backends_probability.values.sum == 1.0
 
+        warn("paths is deprecated, use targets instead") if paths
```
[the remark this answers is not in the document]
bmansoob: I think this should suffice https://github.com/Shopify/app_profiler/pull/134/commits/9507a7e50c17b7385161f42b5be50ec2c8fc916c

lib/app_profiler.rb
```diff
@@ -116,6 +116,25 @@ def backend=(new_backend)
...
+          "profile_sampler_enabled must be a proc or a lambda that accepts no argument" if value.arity != 0
+      else
+        raise ArgumentError, "Must be TrueClass or FalseClass" unless [TrueClass, FalseClass].include?(value.class)
+      end
+
+      @profile_sampler_enabled = value
+    end
+
+    def profile_sampler_enabled
+      return false unless defined?(@profile_sampler_enabled)
```
casperisfine, line 131:
```suggestion
```

You don't need that. If it's not defined then it's `nil`. If it's `nil` it isn't a `Proc`. So you just return `nil` which is close enough to `false`.

lib/app_profiler.rb
casperisfine: Maybe log something?

`Rails.error.report` would be ideal, but not sure we can assume it's there.

I'm just a bit worried we ship a faulty sampler and not notice? I guess we'd notice we have no data coming in?

lib/app_profiler/sampler/config.rb
```diff
@@ -5,28 +5,31 @@
...
+        },
+        paths: nil)
 
         if sample_rate < 0.0 || sample_rate > 1.0
           raise ArgumentError, "sample_rate must be between 0 and 1"
         end
 
         raise ArgumentError, "mode probabilities must sum to 1" unless backends_probability.values.sum == 1.0
 
+        ActiveSupport::Deprecation.warn("passing paths is deprecated, use targets instead") if paths
```
casperisfine, line 29: Hum, that method doesn't exist. https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html

casperisfine (approved):

lib/app_profiler/sampler/config.rb
bmansoob: interesting, maybe it was was available in a previous version -- because tests do pass locally. Will change to `ActiveSupport::Deprecation.new`

bmansoob: done https://github.com/Shopify/app_profiler/pull/134/commits/032840c25c3a16082eef85e977b417cec6ff26a9

lib/app_profiler.rb
bmansoob: If it's okay, I would like to keep it as is. We enforce `True/False/Proc` when setting. So if we read `nil` and then try to set it back, it raises - which does not seem right.

bmansoob:
Going to squash the commits now

bmansoob:
Shipped this version to Core, everything is working as expected.

PR #141, opened 2024-10-18
Validate the backend in params

sam-kayg:
Explicitly handle the case when the backend provided by params is not supported by AppProfiler

### DIFF
--- a/lib/app_profiler/request_parameters.rb
+++ b/lib/app_profiler/request_parameters.rb
@@ -27,11 +27,12 @@ module AppProfiler
     end
 
     def valid?
-      if mode.blank?
+      return false if mode.blank?
+
+      unless valid_backend?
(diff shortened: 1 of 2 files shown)


lib/app_profiler/request_parameters.rb
```diff
@@ -50,6 +51,12 @@ def valid?
       true
     end
 
+    def valid_backend?
+      return true if AppProfiler::Backend::StackprofBackend.name == backend
```
bmansoob, line 55: I know it achieves the same outcome and it is probably a matter of personal preference - but I would rather get rid of `AppProfiler::Backend::StackprofBackend.name == backend` and have something which checks if `backend` is one of the two.  We can define a constant `AVAILABLE_BACKENDS`

PR #147, opened 2024-11-21
Require time zone core_ext instead of class

gmcgibbon:
We use Time.zone, which is a core extension. If it isn't loaded, we may not be able to access the zone correctly.

https://github.com/Shopify/app_profiler/actions/runs/11962066460/job/33349616705


gmcgibbon:
The timeouts appear to be happening on main prior to my changes. 🤔 I don't think this will fix them.

PR #148, opened 2024-12-23
Start process_queue_thread when dead

bmansoob:
Fixes the case when process A starts the background thread and process B is forked from process A. In this case, the old code would not recreate the thread because the thread variable is still there but the thread itself is dead. 

We see this a lot in pitchfork because workers are forked from `mold`, which I believe/understand is selected from a worker too.


manuelfelipe (approved):
thanks for the fix and the troubleshooting efforts. Fix makes sense when considering our forking strategy

PR #150, opened 2025-01-27
fix profile-viewer name

bmansoob:
The gem name is `profile-viewer`, `profile_viewer` errors out.


manuelfelipe (approved):

dalehamel (approved):
My guess is when the PR was originally created,

```
 exec("which", "profile-viewer", silent: true) do
```

Would be successful, as the gem was already installed (probably manually?), and the executable was already on the path.

So, the fact that the gem and executable have different casing of the `-` was not detected.

gmcgibbon:
Thanks folks. Sorry about that!

PR #151, opened 2025-01-29
profile file name proc

bmansoob:
Adds a `profile_file_name` proc. This is useful if we want to make profile names predictable and link to the observability signals, like `trace_id`. README specifies its usage.


lib/app_profiler.rb
```diff
@@ -129,6 +129,14 @@ def vernier_viewer
       @@vernier_viewer ||= Viewer::FirefoxViewer # rubocop:disable Style/ClassVars
     end
 
+    # mattr_reader :profile_file_name
```
manuelfelipe, line 132: nit: want to clean the commented out code?

PR #154, opened 2025-01-29
Reintroduce VernierBackend definition check

KaanOzkan:
We have to safely ensure vernier is available and we can load the constant

### DIFF
--- a/lib/app_profiler.rb
+++ b/lib/app_profiler.rb
@@ -173,7 +173,7 @@ module AppProfiler
     end
 
     def vernier_supported?
-      RUBY_VERSION >= "3.2.1"
+      RUBY_VERSION >= "3.2.1" && defined?(AppProfiler::Backend::VernierBackend.name)
     end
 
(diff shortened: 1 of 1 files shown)


bmansoob (approved):

lib/app_profiler.rb
```diff
@@ -173,7 +173,7 @@ def backend
     end
 
     def vernier_supported?
-      RUBY_VERSION >= "3.2.1"
+      defined?(AppProfiler::Backend::VernierBackend.name) && RUBY_VERSION >= "3.2.1"
```
gmcgibbon, line 176: These conditions should probably be reversed. The version check is likely lighter than the defined? check.

gmcgibbon (approved):

lib/app_profiler.rb
KaanOzkan: I was thinking we should prioritize the one that's more likely to fail. Wdyt?

PR #156, opened 2025-02-03
add profile_id

bmansoob:
Add profile_id generator. The motivation behind this is to make file names predictable using the proc added in https://github.com/Shopify/app_profiler/pull/151.

`Id` is defined in a similar like in [OTEL](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/blob/9205feec476e1d47dfb61731aa40a950eef3c8ef/reporter/internal/pdata/generate.go#L42-L48).


bmansoob:
@gmcgibbon would appreciate your feedback here specifically ruby/rails related stuff.

lib/app_profiler/profile_id.rb
```diff
@@ -0,0 +1,26 @@
...
+require "active_support/current_attributes"
+require "securerandom"
+
+module AppProfiler
+  class ProfileId
+    class Current < ActiveSupport::CurrentAttributes
+      attribute :id
+      def reset
+        self.id = nil
+      end
```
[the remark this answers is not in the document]
bmansoob: thanks -- i was not aware, will remove.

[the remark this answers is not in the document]
bmansoob: done https://github.com/Shopify/app_profiler/pull/156/commits/0adc6b06c9429d7da24330bd8bed192c4c324ba5

lib/app_profiler/profile_id.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    class << self
+      def current
+        Current.id ||= SecureRandom.hex(16)
+        Current.id
+      end
+
+      def reset
+        Current.reset
+      end
```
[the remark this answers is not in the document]
bmansoob: done https://github.com/Shopify/app_profiler/pull/156/commits/0adc6b06c9429d7da24330bd8bed192c4c324ba5

lib/app_profiler/profile_id.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "active_support/current_attributes"
+require "securerandom"
+
+module AppProfiler
+  class ProfileId
```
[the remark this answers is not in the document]
gmcgibbon: Sure, but you only need it for a profile. The current name is already keyed to this concept.

lib/app_profiler/base_profile.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "active_support/deprecation/constant_accessor"
-
```
gmcgibbon, line 4: We should add this back. Requires are usually spaced just above the class body.

lib/app_profiler/base_profile.rb
```diff
@@ -17,15 +16,15 @@ class << self
...
+        StackprofProfile.new(data, id: ProfileId.current, **options).tap do |profile|
           raise ArgumentError, "invalid profile data" unless profile.valid?
         end
       end
 
       def from_vernier(data)
         options = INTERNAL_METADATA_KEYS.map { |key| [key, data[:meta]&.delete(key)] }.to_h
 
-        VernierProfile.new(data, **options).tap do |profile|
+        VernierProfile.new(data, id: ProfileId.current, **options).tap do |profile|
```
gmcgibbon, line 27: The profile ID would be wrong if the ID option is passed in. If the ID option is passed in, we should probably assign the ID, right?

test/app_profiler/profile_id_test.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module AppProfiler
+  class ProfileIdTest < TestCase
+    test "ProfileId memoization" do
```
gmcgibbon, line 7:
```suggestion
    test ".current memoization" do
```

test/app_profiler/profile_id_test.rb
```diff
@@ -0,0 +1,16 @@
...
+      ProfileId.reset
+
+      SecureRandom.expects(:hex).once.returns("profile-id")
```
gmcgibbon, line 10:
```suggestion
```
Don't need to mock for this test. Part of this test would be to actually call the secure random call because we mock it in most other places. 😄

test/app_profiler/profile_id_test.rb
```diff
@@ -0,0 +1,16 @@
...
+      id = ProfileId.current
+      assert_equal(id, ProfileId.current)
+      assert_equal(id, ProfileId.current)
```
gmcgibbon, line 13:
```suggestion
      assert_same(id, ProfileId.current)
```
And then remove the first equal assert.

lib/app_profiler/profile_id.rb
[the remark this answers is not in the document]
gmcgibbon: The method is still there.

lib/app_profiler/base_profile.rb
bmansoob: you are right -- i guess we would like to keep the existing behaviour intact. I have changed this -- so when its not supplied we will fallback

`@id  = id.presence || ProfileId.current`

test/app_profiler/profile_id_test.rb
bmansoob: good feedback 👍 -- better than testing it via expectations

gmcgibbon (approved):
LGTM once `ProfileId.reset` is cleaned up. Thanks! Feel free to squash your commits too so the change is more concise.

PR #161, opened 2025-02-19
add otel instrumentation

bmansoob:
We recently integrated profiling with tracing and the feedback we have gotten is quite encouraging. 

The ability to see traces and profiles side by side is something internal users have asked for sometime now. To that end, It makes sense to integrate tracing and profiling here in the gem - to avoid repetitive rework in clients.

This PR adds a new `otel_instrumentation_enabled` as a new config option. It can only be enabled if `OpenTelemetry::Instrumentation::Rack` is loaded. `app_profiler` does not require this explicitly, it is optional. 

All our apps use `otel` gems - so things should work out of the box for them.


robertlaurin (approved):

lib/app_profiler.rb
```diff
@@ -162,6 +164,14 @@ def profile_sampler_enabled
       false
     end
 
+    def otel_instrumentation_enabled=(value)
+      if value
+        raise ArgumentError,
```
bmansoob, line 169: I will relax this condition a bit. Rather than raising - will set it to false and warn if otel is not defined.

bmansoob:
@gmcgibbon if you could please take a look from ruby/rails perspective. Much appreciated 🙏

test/app_profiler/app_profiler_configuration_test.rb
```diff
@@ -50,5 +50,13 @@ class ConfigurationTest < TestCase
     ensure
       AppProfiler.profile_sampler_enabled = old_status
     end
+
+    test "otel_instrumentation_enabled raises ArgumentError if opentelemetry is not installed" do
```
bmansoob, line 54: will update the description since we are no longer raising.

gmcgibbon: Hmmmm... With the approach I'm suggesting, we can likely mock the `gem` call to raise `Gem::MissingSpecError` instead.

lib/app_profiler.rb
```diff
@@ -162,6 +164,17 @@ def profile_sampler_enabled
       false
     end
 
+    def otel_instrumentation_enabled=(value)
+      if value
+        unless defined?(OpenTelemetry::Instrumentation::Rack)
+          logger.warn("OpenTelemetry::Instrumentation::Rack not defined. Setting enabled to false.")
+          @otel_instrumentation_enabled = false
+          return
+        end
```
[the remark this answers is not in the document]
bmansoob: clarified privately -- I understand the suggestion now.

test/app_profiler/app_profiler_configuration_test.rb
bmansoob: Mocking does not work -- because `gem` is monkey patched here? https://github.com/rubygems/rubygems/blob/bundler-v2.5.9/bundler/lib/bundler/rubygems_integration.rb#L206. 

I can remove the test altogether and just rescue `Gem::MissingSpecError` and `Gem::LoadError` in code and leave at that.

lib/app_profiler/middleware.rb
```diff
@@ -47,6 +54,31 @@ def profile(env, params)
...
+
+      attributes = {
+        OTEL_PROFILE_ID => profile_id,
+        OTEL_PROFILE_BACKEND => params[:backend].to_s,
+        OTEL_PROFILE_MODE => params[:mode].to_s,
+        OTEL_PROFILE_CONTEXT => AppProfiler.context,
+      }
+
+      rack_span.add_attributes(attributes)
+    rescue => e
```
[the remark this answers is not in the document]
bmansoob: I was a bit conflicted when adding this -- just being paranoid about failing the request in case something breaks here. I dug a bit into the gem --  I can just remove this altogether.

test/app_profiler/middleware_test.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require "test_helper"
+require "opentelemetry/instrumentation/rack"
+require "opentelemetry/sdk"
```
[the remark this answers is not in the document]
bmansoob: removed https://github.com/Shopify/app_profiler/pull/161/commits/fbe1e772cac66f19bf7bb049d9c68b0558734d3b#diff-734be06f68efb285ac767da13d7ff44b8adc7e093a4a33cd0b4814b772248530L38

lib/app_profiler/middleware.rb
```diff
@@ -7,6 +7,11 @@
 
 module AppProfiler
   class Middleware
+    OTEL_PROFILE_ID = "profile.id"
```
bmansoob, line 10: https://github.com/open-telemetry/semantic-conventions/blob/d963d60dc076d80d630acadcdf265c2bfb17e8a1/model/profile/common.yaml

lib/app_profiler.rb
```diff
@@ -162,6 +164,15 @@ def profile_sampler_enabled
       false
     end
 
+    def otel_instrumentation_enabled=(value)
+      require("opentelemetry/instrumentation/rack") if value
```
gmcgibbon, line 168:
```suggestion
      if value
        gem("opentelemetry-instrumentation-rack")
        require("opentelemetry/instrumentation/rack")
      end
```

lib/app_profiler.rb
```diff
@@ -162,6 +164,15 @@ def profile_sampler_enabled
...
+
+      @otel_instrumentation_enabled = value
+    rescue Gem::LoadError => e
```
gmcgibbon, line 171: Since we want the program to crash if this is enabled and we don't have the gem, I don't think this is necessary to rescue.

lib/app_profiler.rb
bmansoob: I am not sure if I understand the comment here 🤔 

>A log entry could easily be missed, approved, and deployed. We don't edit Rails config after boot, so this should be safe to do.

RE: `gem("opentelemetry-instrumentation-rack")` 👍 

>  If you will be activating the latest version of a gem, there is no need to call Kernel#gem, Kernel#require will do the right thing for you.

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/core_ext/kernel_gem.rb#L11-L12. Do we still need to call `gem`?

lib/app_profiler.rb
bmansoob: I was preferring not to crash the program -- since we may want to enable it for all our apps via our metrics gem. Some app out in the wild may not have the otel gem for some reason and we do not want it to fail on boot.

Also, design-wise, since this is just instrumentation I think it is okay to rescue silently and disable it.

lib/app_profiler.rb
bmansoob: 👍

gmcgibbon: > I am not sure if I understand the comment here 🤔

You're rescuing the gem missing error, and logging that the gem doesn't exist. That's not helpful in instances where the developer thinks they have the right gem but doesn't. It is far easier to miss. IMO the app should crash with a clear error message because it requires intervention from the developer to add the right gem.

bmansoob: done and done

lib/app_profiler.rb
bmansoob: crash it is 👍

bmansoob:
I will squash

bmansoob:
... and verified that everything still works for our app.

```
#<OpenTelemetry::SDK::Trace::Span:0x00007edd01170c98
 @attributes=
  {....
   "profile.id" => "3e0783dff80528b19236b298fd0906ef",
   "profile.profiler" => "stackprof",
   "profile.mode" => "cpu",
   "profile.context" => "XXXX"},
```

PR #163, opened 2025-02-28
insert profiling middleware after Rack::Events

bmansoob:
I believe `Rack::Events` is usually one of the first middlewares in chain. For Core, it is the first. 

But for some apps it seems like it can appear after the profiling middleware. In that scenario, otel tracing attributes would not work.

Is this okay to do? @gmcgibbon

### DIFF
--- a/lib/app_profiler/railtie.rb
+++ b/lib/app_profiler/railtie.rb
@@ -60,7 +60,11 @@ module AppProfiler
         if (Rails.env.development? || Rails.env.test?) && AppProfiler.stackprof_viewer.remote?
           app.middleware.insert_before(0, AppProfiler.viewer::Middleware)
         end
-        app.middleware.insert_before(0, AppProfiler.middleware)
+        if AppProfiler.otel_instrumentation_enabled
+          app.middleware.insert_after(Rack::Events, AppProfiler.middleware)
+        else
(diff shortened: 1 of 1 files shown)


lib/app_profiler/railtie.rb
```diff
@@ -60,7 +60,7 @@ class Railtie < Rails::Railtie
         if (Rails.env.development? || Rails.env.test?) && AppProfiler.stackprof_viewer.remote?
           app.middleware.insert_before(0, AppProfiler.viewer::Middleware)
```
dalehamel, line 61: i think we may need to ensure that this is inserted after rack events too, otherwise this middleware may be before the main one?

lib/app_profiler/railtie.rb
```diff
@@ -60,7 +60,11 @@ class Railtie < Rails::Railtie
...
         end
-        app.middleware.insert_before(0, AppProfiler.middleware)
+        if AppProfiler.otel_instrumentation_enabled
```
bmansoob, line 63: Otel gems use `Rack::Events`. So inserting the profiling middleware, after `Rack::Events`, based on the config value makes sense to me.

What do you think @gmcgibbon?

gmcgibbon (approved):

PR #165, opened 2025-04-03
Support vernier metadata

dalehamel:
# What

Move's metadata storage into a namespace under "meta" within vernier profiles, and pass "metadata" param directly to the collector on versions >= 1.7.0

# Why

Vernier 1.7.0 adds support for metadata as part of the output natively thanks to these two PRs:

https://github.com/jhawthorn/vernier/pull/140
https://github.com/jhawthorn/vernier/pull/145

With these, metadata specified in `vernier run` or via ruby API will be stored in two places:

- meta.vernierUserMetadata
  - easily machine readable
- meta.extra
  - For the UI to display

# How

This moves the metadata accessor for vernier profiles to be namespaced under `vernierUserMetadata` key. For compatibility with older versions of vernier, if this key does not exist metadata is copied here.

We also have some "internal" metadata that vernier adds to its "result" object, but does not include in the output json. Since this data is potentially useful, we copy this as if it were user specified. In the case that user specified metadata has the same key, the "internal" one takes precedence. The exception is the `started_at` metadata, which would be useful if it were in realtime, but as it is from a monotonic clock source it is not especially helpful outside of vernier internals.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gemspec
 # Specify the same dependency sources as the application Gemfile
 gem("activesupport", "~> 5.2")
 gem("railties", "~> 5.2")
-gem("vernier", "~> 1.3.1") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
+gem("vernier", "~> 1.7.0") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
 
 gem("google-cloud-storage", "~> 1.21")
 gem("rubocop", require: false)
--- a/lib/app_profiler/backend/vernier_backend.rb
+++ b/lib/app_profiler/backend/vernier_backend.rb
@@ -76,6 +76,14 @@ module AppProfiler
 
         return unless vernier_profile
 
+        # Store all vernier metadata
+        meta = vernier_profile.meta.reject { |k, v| k == :user_metadata || v.nil? }
+        meta.merge!(@metadata) if @metadata
+
+        # Internal metadata takes precedence over user metadata, but store
+        # everything in user metadata
+        vernier_profile.meta[:user_metadata]&.merge!(meta)
+
         # HACK: - "data" is private, but we want to avoid serializing to JSON then
         # parsing back from JSON by just directly getting the hash
         data = ::Vernier::Output::Firefox.new(vernier_profile).send(:data)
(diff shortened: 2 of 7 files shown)


lib/app_profiler/backend/vernier_backend.rb
```diff
@@ -43,7 +43,9 @@ def start(params = {})
         @mode = params.delete(:mode) || DEFAULTS[:mode]
         raise ArgumentError unless AVAILABLE_MODES.include?(@mode)
 
-        @metadata = params.delete(:metadata)
+        if Gem.loaded_specs["vernier"].version < Gem::Version.new("1.7.0")
```
dalehamel, line 46: In 1.7.0 or greater, we can just pass it directly as as it was added in https://github.com/jhawthorn/vernier/pull/145/files

lib/app_profiler/backend/vernier_backend.rb
```diff
@@ -74,11 +76,19 @@ def results
...
+        # Internal metadata takes precedence over user metadata, but store
+        # everything in user metadata
+        vernier_profile.meta[:user_metadata]&.merge!(meta)
+
         # HACK: - "data" is private, but we want to avoid serializing to JSON then
         # parsing back from JSON by just directly getting the hash
         data = ::Vernier::Output::Firefox.new(vernier_profile).send(:data)
         data[:meta][:mode] = @mode # TODO: https://github.com/jhawthorn/vernier/issues/30
-        data[:meta].merge!(@metadata) if @metadata
+        data[:meta][:vernierUserMetadata] ||= meta # for compatibility with < 1.7.0
```
dalehamel, line 91: for metadata, we now read directly from this output key. To be compatible with older versions, we'll store the metadata that was added by the user here

Gemfile
```diff
@@ -6,7 +6,7 @@ gemspec
 # Specify the same dependency sources as the application Gemfile
 gem("activesupport", "~> 5.2")
 gem("railties", "~> 5.2")
-gem("vernier", "~> 1.3.1") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
+gem("vernier", "~> 1.7.0") if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.2.1")
```
dalehamel, line 9: Bump this for local development of vernier, so that all tests will run.

test/app_profiler/backend/vernier_backend_test.rb
```diff
@@ -77,7 +77,7 @@ def teardown
         assert_instance_of(AppProfiler::VernierProfile, profile)
         assert_equal("wowza", profile.id)
         assert_equal("bar", profile.context)
-        assert_equal("spam", profile[:meta][:extrameta])
+        assert_equal("spam", profile.metadata[:extrameta])
```
dalehamel, line 80: This was already a bit leaky, we should have been accessing it through the metadata helper method.

bmansoob (approved):

PR #210, opened 2026-04-28
Require Ruby 3.2.1+

manuelfelipe:
## What

Raise AppProfiler's supported Ruby floor to Ruby 3.2.1, test the currently relevant Ruby 3.x line (3.2, 3.3, and 3.4), and bump the gem to 0.6.0 for the support-policy change.

## Why

AppProfiler is a runtime/profiling integration, not a plain Ruby utility gem. Keeping very old Ruby versions working is increasingly not a passive compatibility promise: it requires different dependency graphs, conditionals in the development bundle, and CI-only fallback behavior for profiler/runtime gems that follow modern Ruby internals.

The versions being dropped here are also outside upstream Ruby security support:

| Ruby | Released | Upstream EOL | Current state |
| --- | --- | --- | --- |
| 2.7 | December 2019 | March 2023 | ~3 years past EOL |
| 3.0 | December 2020 | April 2024 | ~2 years past EOL |
| 3.1 | December 2021 | March 2025 | ~1 year past EOL |

So this is not a move to "latest Ruby only". It removes compatibility paths for Rubies that are already past upstream security support while keeping AppProfiler on a Ruby 3.2+ baseline.

Today legacy support is split across the normal `Gemfile` and `Gemfile.legacy-rubies`. That has a few practical costs:

- CI, local development, releases, and automated dependency tooling are not all exercising the same bundle.
- Dependabot/Shipit security-update flows use the default dependency files, not the CI fallback Gemfile, so old-Ruby compatibility can break release automation even when CI has a special path.
- Dependency updates become harder to reason about because modern profiler dependencies need to be guarded or excluded for old Rubies.
- The value of continuing to support Ruby 2.7/3.0/3.1 is unclear now that these are legacy runtimes, while the maintenance cost keeps increasing.

The recent Vernier metadata work is the thing that exposed the problem, but it is not the only reason for this change. Vernier is part of the broader signal: the profiler ecosystem is moving with newer Rubies, and AppProfiler's support matrix should match the Ruby versions we are willing to actively test and release against.

Rather than keep extending the split-brain setup, this PR makes the supported surface explicit and keeps CI focused on Ruby 3.2+.

## Why 3.2.1?

Ruby 3.2.1 is the smallest floor that works with the current profiling stack after the Vernier updates. Vernier requires Ruby `>= 3.2.1`, and AppProfiler already had Vernier-specific docs/guards around that boundary.

This keeps support broader than "latest Ruby only" while removing the legacy-only dependency path.

## Changes

- Set `spec.required_ruby_version` to `>= 3.2.1`
- Remove the Ruby-version guard around the development `vernier` dependency
- Remove `Gemfile.legacy-rubies`
- Drop CI jobs for Ruby 2.7, 3.0, and 3.1
- Add CI test coverage for Ruby 3.4, so the test matrix covers 3.2, 3.3, and 3.4
- Add explicit `base64`, `bigdecimal`, and `mutex_m` dependencies needed by ActiveSupport 5.2 under Ruby 3.4/Bundler
- Document the Ruby 3.2.1+ requirement
- Bump the gem to 0.6.0 and update the changelog

## Why not Ruby 4.0 yet?

Ruby 4.0 exposed older development dependency constraints in the current Rails 5.2 / google-cloud-storage / Faraday graph. Keeping Ruby 4.0 out of this PR avoids turning a support-policy cleanup into a broader dependency modernization.

Ruby 4 support is worth doing separately once we are ready to modernize that dependency graph deliberately.

### Keep Ruby 2.7/3.0/3.1 support

We could preserve `Gemfile.legacy-rubies` and add a Dependabot/Shipit-specific workaround. That would avoid a support-policy change, but it would also keep the default bundle, CI fallback bundle, and release automation pointed at different dependency realities.

### Keep conditional Vernier support only

We could continue guarding Vernier and treating old Ruby support as best-effort. That still leaves the release/dependency tooling problem and makes future profiler work more likely to require more conditionals.

## Testing

- `shadowenv exec -- bundle exec rake test`
- `shadowenv exec -- bundle exec rubocop`
- Ruby 3.4 local check: `BUNDLE_PATH=/tmp/app_profiler_bundle_34 /opt/rubies/3.4.8/bin/bundle exec rake test`
- `git diff --check`

## Discussion

This is a support-policy change. If we want to keep Ruby 2.7/3.0/3.1 support, we should explicitly commit to maintaining the legacy bundle path and add a dedicated workaround for Dependabot/Shipit rather than relying on CI fallback behavior alone.


dalehamel (approved):
Looks good, let's just do a non-patch level bump to 0.6.0 since we're dropping support.

PR #217, opened 2026-07-12
Skip middleware actions for zero-sample profiles

### LINKED ISSUE #2: Middleware's action should not be called if there are no samples in the profile
We should avoid calling the middleware's action (e.g. UploadAction) if there are no samples in the profile.

https://github.com/Shopify/app_profiler/blob/80df44c69c8bd7dedb1c250c1ecec422ffa4f838/lib/app_profiler/middleware.rb#L41-L45

When a request returns quickly during a profiling session, there is a possibility in which the profiler could not sample anything in time. There won't be any samples for those profiles, and speedscope won't recognize them either. We should drop such profiles.

One way to do so would be to return `nil` if the value for `samples` is 0 whenever profiling results are retrieved:
https://github.com/Shopify/app_profiler/blob/80df44c69c8bd7dedb1c250c1ecec422ffa4f838/lib/app_profiler/profiler.rb#L30

oiahoon:
## Summary

**Unusable empty profiles no longer trigger middleware actions.** The middleware now returns the original response when StackProf reports zero samples instead of uploading, enqueueing, or viewing a profile that Speedscope cannot consume.

## Behavior

The zero-sample guard runs after `after_profile`, preserving the existing hook behavior and direct backend/profile APIs. Profiles with positive samples and backends without the StackProf `samples` field are unchanged.

## Verification

- failure-before regression reproduced the unexpected `UploadAction.call`
- middleware tests passed with three sequential seeds: 47 runs and 281 assertions each
- full Ruby 3.4 suite: 250 runs, 836 assertions
- RuboCop: 70 files, no offenses
- gem build
- `git diff --check`

Closes #2


oiahoon:
I have signed the CLA!

PR #218, opened 2026-09-14
Honor disabled CORS in Rails configuration

dalehamel:
## Summary

The Rails initializer currently applies `server_cors || true`, which discards an explicitly configured `false`. Preserve `false` while keeping CORS enabled when the option is unset or `nil`.

- Add isolated regression tests that run the actual `app_profiler.configs` initializer and check successful `/profile` response headers for unset, `nil`, `true`, `false`, and custom-origin configurations.
- Document the Rails CORS opt-out and origin restriction, and clarify that localhost binding and CORS are not authentication.
- Refresh the repository ownership metadata.

## Validation

- Both explicit-disable regression cases fail against the previous initializer and pass with this fix.
- `bundle exec rake`: 255 tests, 851 assertions, no failures or errors (Ruby 3.3.1); also passed with five fixed random seeds.
- `bundle exec rubocop`: 71 files inspected, no offenses.


manuelfelipe (approved):
thanks. fine for now to address the linked issue. 

in any case, as discussed, i think we should just look into fully deprecating and removing the server support as we no longer make use of it

bmansoob (approved):

